PR #15, opened 2023-03-12
Podman (tested in Fedora 37)

borislassort:
This is a small patch to use podman (by default in Fédora) when it's available.
And `unshare chown $UID:$UID` otherwise the container could not write in ~/.zwift/

### DIFF
--- a/zwift.sh
+++ b/zwift.sh
@@ -1,7 +1,8 @@
 #!/bin/bash +x
 
 # Use podman if available
-if [ -x "$(command -v podman)" ]; then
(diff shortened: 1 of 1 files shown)


zwift.sh
```diff
@@ -1,4 +1,12 @@
-#!/bin/bash
+#!/bin/bash +x
+
+# Use podman if available
+if [[ -x "$(command -v podman)" ]]
```
netbrain, line 4: What if podman and docker is available? What should be the default course of action? Prefer podman?

netbrain (approved):

PR #16, opened 2023-07-17
Use xhost on non-Wayland systems only

quietvoid:
1. On my system using `xhost` is not necessary at all. I don't think it's exclusive to `XWayland` though.
I've made it run only when not running Wayland but I don't think it's the "correct" thing to do.
3. `docker` does not have a `unshare` command, so I moved it to the `podman` condition.

Unfortunately I wasn't able to properly test on my side with `podman`.
I keep running into the container having issues accessing the config volume, regardless of rootless or not.


netbrain:
Looks good to me. Thank you for this!

PR #18, opened 2023-09-03
Add support for autologin into Zwift game

### LINKED ISSUE #17: Zwift password
Hi.
Every time I start zwift docker the app welcomes me with login dialog, where email is already entered, but password field is blank.

I'd like to ask if there is an option to make zwift docker auto login to my account or just not forgetting password :)

quietvoid:
Fixes #17.

Todo:
- [x] `curl` should probably be installed in the parent image (`netbrain/wine`), otherwise we have to pull sources again for no reason.
- [x] Figure out how to fix `RunFromProcess` truncating the JSON token, resulting in autologin failure.
I've contacted NirSoft FWIW, will see if there's a workaround or a required fix in the program.

There's also the alternative of switching to https://github.com/quietvoid/runfromprocess-rs but it needs more testing.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -8,15 +8,19 @@ mkdir -p "$ZWIFT_HOME"
 cd "$ZWIFT_HOME"
 
 function get_current_version() {
-	ZWIFT_VERSION_CURRENT=$(cat Zwift_ver_cur.xml | grep -oP 'sversion="\K.*?(?=")' | cut -f 1 -d ' ')
+	if grep -q sversion Zwift_ver_cur.xml; then
+		ZWIFT_VERSION_CURRENT=$(cat Zwift_ver_cur.xml | grep -oP 'sversion="\K.*?(?=")' | cut -f 1 -d ' ')
+	else
+		# Basic install only, needs initial update
+		ZWIFT_VERSION_CURRENT="0.0.0"
+	fi
 }
 
 function get_latest_version() {
 	ZWIFT_VERSION_LATEST=$(wget --quiet -O - http://cdn.zwift.com/gameassets/Zwift_Updates_Root/Zwift_ver_cur.xml | grep -oP 'sversion="\K.*?(?=")' | cut -f 1 -d ' ')
 }
 
-if [ "$1" = "update" ]
-then
+function wait_for_zwift_game_update() {
 	echo "updating zwift..."
 	get_current_version
 	get_latest_version
@@ -25,15 +29,23 @@ then
 		echo "already at latest version..."
 		exit 0
 	fi
-	wine64 start ZwiftLauncher.exe
+
+	wine64 start ZwiftLauncher.exe SilentLaunch
 	until [ "$ZWIFT_VERSION_CURRENT" = "$ZWIFT_VERSION_LATEST" ]
 	do
 		echo "updating in progress..."
 		sleep 5
 		get_current_version
 	done
+
 	echo "updating done, waiting 5 seconds..."
 	sleep 5
+}
+
+if [ "$1" = "update" ]
+then
+	wait_for_zwift_game_update
+
 	wineserver -k
 	exit 0
 fi
@@ -48,10 +60,6 @@ then
 
 	#workaround crash issue 1.21
 	winetricks --unattended d3dcompiler_47
-	
-	#install msedge
-	wget https://msedge.sf.dl.delivery.mp.microsoft.com/filestreamingservice/files/efb9eb0d-607c-4fd7-9304-0f5c3ca433fb/MicrosoftEdgeWebView2RuntimeInstallerX64.exe
-	wine64 MicrosoftEdgeWebview2RuntimeInstallerX64.exe /silent /install
 
 	#install zwift
 	wget https://www.nirsoft.net/utils/runfromprocess.zip
@@ -65,23 +73,40 @@ then
 	do
 		echo "updating in progress..."
 		sleep 5
-
 	done
+
+	# Restart updates in background
+	pkill ZwiftLauncher || true
+	wait_for_zwift_game_update
+
 	wineserver -k
 	exit 0
 fi
+
 echo "starting zwift..."
-wine64 start ZwiftLauncher.exe
-wine64 start RunFromProcess-x64.exe ZwiftLauncher.exe ZwiftApp.exe
+wine64 start ZwiftLauncher.exe SilentLaunch
+
+if [[ -f "/home/user/Zwift/.zwift-credentials" ]]
+then
+	LAUNCHER_PID_HEX=$(winedbg --command "info proc" | grep -P "ZwiftLauncher.exe" | grep -oP "^\s\K.+?(?=\s)")
+	LAUNCHER_PID=$((16#$LAUNCHER_PID_HEX))
+
+	echo "authenticating with zwift..."
+	wine64 start /exec /bin/runfromprocess-rs.exe $LAUNCHER_PID ZwiftApp.exe --token=$(zwift-auth)
+
+	sleep 3
+else
+	wine64 start RunFromProcess-x64.exe ZwiftLauncher.exe ZwiftApp.exe 
+fi
+
 until pgrep ZwiftApp.exe &> /dev/null
 do
-    echo "Waiting for zwift to start ..."
-    sleep 1
+	echo "Waiting for zwift to start ..."
+	sleep 1
 done
 
 echo "Killing uneccesary applications"
 pkill ZwiftLauncher
-pkill MicrosoftEdgeUp
 pkill ZwiftWindowsCra
 
 wineserver -w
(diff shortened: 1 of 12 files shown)


entrypoint.sh
```diff
@@ -70,9 +70,18 @@ then
...
 fi
+
 echo "starting zwift..."
 wine64 start ZwiftLauncher.exe
-wine64 start RunFromProcess-x64.exe ZwiftLauncher.exe ZwiftApp.exe
+
+if [[ -f "/home/user/Zwift/.zwift-credentials" ]]
+then
+	echo "authenticating with zwift..."
+	wine64 start RunFromProcess-x64.exe ZwiftLauncher.exe ZwiftApp.exe --token=$(zwift-auth)
```
SirYaro: Shouldn't it be _$(zwift-auth.sh)_ here?

quietvoid: The script is renamed when copying in the image.

quietvoid:
Current status is that it seems to work just barely with these diffs and the `runfromprocess-rs.exe` here:
[runfromprocess-rs.zip](https://github.com/netbrain/zwift/files/12508296/runfromprocess-rs.zip)



```diff
diff --git a/Dockerfile b/Dockerfile
index 51d151f..a41b331 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,4 +15,7 @@ RUN sudo chmod +x /bin/entrypoint
 COPY zwift-auth.sh /bin/zwift-auth
 RUN sudo chmod +x /bin/zwift-auth
 
+COPY runfromprocess-rs.exe /bin/runfromprocess-rs.exe
+RUN sudo chmod +x /bin/runfromprocess-rs.exe
+
 ENTRYPOINT ["entrypoint"]
```

```diff
diff --git a/entrypoint.sh b/entrypoint.sh
index ddfd9c2..95fd1fe 100644
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -76,17 +76,26 @@ wine64 start ZwiftLauncher.exe
 
 if [[ -f "/home/user/Zwift/.zwift-credentials" ]]
 then
+       LAUNCHER_PID_HEX=$(winedbg --command "info proc" | grep -P "ZwiftLauncher.exe" | grep -oP "^\s\K.+?(?=\s)")
+       LAUNCHER_PID=$((16#$LAUNCHER_PID_HEX))
+
        echo "authenticating with zwift..."
-       wine64 start RunFromProcess-x64.exe ZwiftLauncher.exe ZwiftApp.exe --token=$(zwift-auth)
+       wine64 /bin/runfromprocess-rs.exe $LAUNCHER_PID ZwiftApp.exe --token=$(zwift-auth)
+
+       until winedbg --command "info proc" | grep "ZwiftApp.exe" &> /dev/null
+       do
+               echo "Waiting for zwift to start ..."
+               sleep 1
+       done
 else
-       wine64 start RunFromProcess-x64.exe ZwiftLauncher.exe ZwiftApp.exe
-fi
+       wine64 start RunFromProcess-x64.exe ZwiftLauncher.exe ZwiftApp.exe 
 
-until pgrep ZwiftApp.exe &> /dev/null
-do
-    echo "Waiting for zwift to start ..."
-    sleep 1
-done
+       until pgrep ZwiftApp.exe &> /dev/null
+       do
+               echo "Waiting for zwift to start ..."
+               sleep 1
+       done
+fi
 
 echo "Killing uneccesary applications"
 pkill ZwiftLauncher
```

No idea how stable the game actually is afterwards.

netbrain:
Awesomeness. Should your rust process also be part of this repo? And we'll
make a GitHub action to build it and incorporate it into the final image?

zwift-auth.sh
netbrain, line 1: This file could probably be a part of the entrypoint.sh so that it's called when the argument "auth" is given, much in the same we do "update" in order to update the image.

So this would be a one time operation in order to create the auth token, and store it on a data volume.

And on the next invocation of entrypoint, if it detects the zwift credentials, then it simply uses them, if not. just launch normally where you will be prompted for login within the game.

zwift.sh should then probably also be modified to accepting "auth" argument and in that case forwarding the argument to `docker run .... netbrain/zwift auth`

netbrain: The general workflow for the user would then be:
1. download `zwift.sh`
2. execute `zwift.sh auth`

Next time auth won't be necessary and can simply call zwift.sh without args.

quietvoid: The auth token has an expiration though.
I don't feel like managing it, so I just went with authenticating on launch.

netbrain:
Ah yes of course. In that case it needs to be run at every launch.

quietvoid:
Did a short 45min test ride and no issue on my side with the changes in the PR.

I think it just needs to get rid of the `until winedbg --command "info proc" | grep "ZwiftApp.exe" &> /dev/null` and just use a fixed sleep time.
It doesn't really matter since it's just closing the launcher, but if it's too early the game gets killed as well.
So a sleep makes sure the window is actually open.

netbrain:
Have to add curl  to the current Dockerfile and then rebuild zwift from scratch i'm afraid. Even if we were to do this in the wine image, this would not change the fact that the docker image has to be rebuilt.

An alternative is to create a new image from a container instance after doing the necessary changes. However i would prefer it to be reproducible by adding all required dependencies (including your rust code) into the original Dockerfile, even if it means we have to download zwift all over again.

Have you tried to build the docker image from scratch using the instructions in the README.

netbrain:
Regarding the rust code, we can easily store it in a subdirectory of this repo, and have a githubaction that builds it and packages  (that builds all from scratch)  it into a new image.

quietvoid:
> Have you tried to build the docker image from scratch using the instructions in the README.

Yes, the WebView2 installer URL is a 404.
Why did it download anyways? There's a bundled installer that comes with Zwift.

I'll try testing with latest/bundled but it didn't work previously when I tried, although with Wine 8.

netbrain:
Would like to help you out with this, but I don't really have the time
right now. But I'll assist with what I can.

quietvoid:
It would be simple to update the new image and commit it, but I don't think rebuilding from scratch is possible currently.
Maybe if there's a way to figure out the same WebView2 version and find the installer for it, maybe it would work.

netbrain:
Yeah, seems the scratch route has issues:
[ERROR] Curl error: [77] 'Problem with the SSL CA cert (path? access rights?)' for: GET https://us-or-rly101.zwift.com/api/telemetry/config
[ERROR] Failed to get telemetry config

Seems zwift has it's own curl implementation and it's not using the correct ca authorities within wine.

quietvoid:
I don't think that's the main issue though.
I'm not able to install either MS Edge or the WebView2 runtime at all, so the launcher never launches.

quietvoid:
Went on Windows, looked up the the startup registry key for Zwift and found that the launcher accepts `SilentLaunch` as argument to start in tray.

I've updated the PR with changes that make building from scratch work.

Tested image build on my side, the updating starts automatically in the background.
It still needs manual intervention for building as you have to click when the Zwift installer asks to download WebView2.
Updating should all be automatic.

I think the image size is even smaller with a clean build compared to current.

Just need to add building `runfromprocess-rs` in the Dockerfile. I'll try to get around to it tomorrow.

netbrain:
Hope you don't mind, but i made some changes. Added the runfromprocess-rs as a part of the build step.

and improved some general script stuff.

Nice catch on the install flag, seems to work like a charm. Good job!

Only thing were missing now is probably some better documentation on the changes and how to get started using autologin feature in a safe manner (i.e chmod zwift-credendtials to 0700) 

Oh, and i've changed the config dir to $HOME/.config/zwift instead of $HOME/.zwift. it's more in the spirit of linux.

And were missing a github action that we can use to create a new release from scratch (i.e. the build-image.sh script)

zwift-auth.sh
```diff
@@ -0,0 +1,64 @@
...
+#     "$LAUNCHER_HOME/ride" \
+#     --write-out "%{url_effective}" | grep -oP "code=zwift_refresh_token\K.+$")
+
+# AUTH_TOKEN_JSON=$(curl -sS --cookie "$COOKIE" --cookie-jar "$COOKIE" \
+#   --data-urlencode "client_id=$LAUNCHER_CLIENT_ID" \
+#   --data-urlencode "redirect_uri=$LAUNCHER_HOME" \
+#   --data-urlencode "refresh_token=$REFRESH_TOKEN" \
+#   --data-urlencode "grant_type=refresh_token" \
+#   --data-urlencode "scope=openid" \
+#   "$ZWIFT_REALM_URL/protocol/openid-connect/token")
```
netbrain, line 60: Leave them as is or remove? are they of any interest to keep?

quietvoid: Depends if one day we want to keep the refresh token around to use as "remember me".
Technically we should only need the user/password once, and then only keep the token saved AFAIK.
At least until the token expires, then it needs to relogin.

Otherwise it's a note to describe how the launcher works.

netbrain (approved):

PR #20, opened 2023-09-13
Using named volume instead of file mounts

### LINKED ISSUE #19: Wayland and podman
Seems the --userns keep-id flag is required for podman to work under wayland, atleast on my test setup.

The following command gave me a proper zwift window

`podman run -d --rm --privileged -e "DISPLAY" -v /tmp/.X11-unix:/tmp/.X11-unix -v /run/user/1000/pulse:/run/user/1000/pulse -v /home/netbrain/.config/zwift/netbrain:/home/user/Zwift --gpus all --userns keep-id docker.io/netbrain/zwift:latest`

Not sure why this is required tho.

Can anyone else confirm this aswell?

netbrain:
Circumventing podman and UID/GID's issues by using a named volume for zwift persisted config instead of relying on file mounts.

Fixes #19

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,8 @@ If you want a proper setup then please use `zwift.sh`.
 
 ## Logging in automatically with zwift.sh
 
-To authenticate through Zwift automatically, the credentials must be present in the mounted persisted Zwift config volume (`$HOME/.config/zwift/$USER/.zwift-credentials`).
+To authenticate through Zwift automatically, the credentials must be present in the zwift volume.
+`docker volume ls` or `podman volume ls` should have a `zwift-$USER` volume.
 A file named `.zwift-credentials` must contain the following lines:
 
 ```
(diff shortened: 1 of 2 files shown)


quietvoid:
It would be nice to have instructions in the readme on creating said volume.
I ran into permission issues since my mounted volume config didn't have the right permissions by default.

Just copied from some googling:
```console
docker run -d --rm --name dummy -v zwift-$USER:/data alpine tail -f /dev/null
docker cp ~/.config/zwift/. dummy:/data
```

But the `Zwift` directory owner was `root` by doing this, instead of `user` (UID 1000).

netbrain:
It's already in the readme I believe.

quietvoid:
It isn't, there's just an example on copying the credentials to the volume.

netbrain:
When launching zwift for the first time:

Podman seems to create permissions root:netbrain (root:user-group) for named volumes if they don't exist. However, when mounted into the zwift container, it seems to be re-mapped to user:user (which is the containers standard user account)

So this should be fine.

Docker seems to create the named volume with netbrain:netbrain permissions on the host system and within the container it's re-mapped to user:user

You are not seeing the same behavior on your system I take it?

quietvoid:
I hadn't tried, I didn't know if Docker would automatically create the named volume so I just did it manually.

netbrain:
Ah i see, in that case you would be better off just deleting the volume you created manually and simply let docker/podman create it when launching the container first time. and then afterwards you can add your specific config files as detailed in the README

PR #25, opened 2023-09-17
Use runfromprocess-rs all the time

### LINKED ISSUE #24: Remove usage of RunFromProcess and rely only on the rust equivalent runfromprocess-rs

quietvoid:
Fixes #24

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -63,8 +63,6 @@ then
     winetricks --unattended d3dcompiler_47
 
     #install zwift
-    wget https://www.nirsoft.net/utils/runfromprocess.zip
-    unzip runfromprocess.zip
     wget https://cdn.zwift.com/app/ZwiftSetup.exe
     wine64 ZwiftSetup.exe /SP- /VERYSILENT /SUPPRESSMSGBOXES /NORESTART /NOCANCEL
     sleep 30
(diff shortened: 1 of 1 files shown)


netbrain:
I think we should have a github action in place to update the current image  with the new entrypoint.sh if it has changed.

As per now this will cause a full rebuild :cry: 

Maybe some logic that does the following:

1. Checkout the repo.
2. Downloads the latest zwift container
3. docker run -d -t zwift netbrain/zwift:latest sleep infinity
4. docker cp zwift:/bin/entrypoint /tmp/entrypoint
5. diff entrypoint.sh /tmp/entrypoint
6. if exit code == 0, then the files are equal, if exit code == 1 then continue to next step
7. docker cp entrypoint.sh zwift:/bin/entrypoint
8. docker commit -m "$git message? or something hardcoded" -c "need to change the command" zwift netbrain/zwift:latest
9. docker tag netbrain/zwift:latest netbrain/zwift:$VERSION
10. docker push netbrain/zwift:latest && docker push netbrain/zwift:$VERSION
11. :partying_face: 

Should also look into which files we should to this diff comparison against. Maybe even the runfromprocess-rs if it ever changes.

Thoughts?

netbrain:
Looks like the files required to be checked as of now is the runfromprocess-rs files, zwift-auth.sh and entrypoint.sh

Could probably also reorganize the repo a bit, so that all files used in the container is in it's own folder.

Would make diffing easier i guess.

quietvoid:
It should probably be both sh scripts only.
`runfromprocess-rs` doesn't belong in the repo, and it's unlikely to change.

netbrain:
Could you elaborate on the "doesn't belong in the repo" part? Do you mean
it was a mistake to include the source code?

I'm open for changes, if you provide some argumentation for your stance on
this.

quietvoid:
> Do you mean it was a mistake to include the source code?

Yes, there's no good reason to do that when you can just clone on build.
https://github.com/quietvoid/zwift/commit/cdd7f97c58654a05fddb56b63a102a6825dbc087#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R10

netbrain:
Im fine with that as long as you promise to never delete the repo :)

quietvoid:
The alternative is that you can just fork it, retag and use your fork.
I don't really care.

PR #31, opened 2023-10-14
[ doc ] note for strong passwords

gallais:
Passwords using special characters need escaping. Otherwise authentication just fails because the variable only got assigned part of the password.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ ZWIFT_PASSWORD=password
 ```
 
 where `username` is your Zwift account email, and `password` your Zwift account password, respectively.
+Note that if your password contains special characters such as `;` you will need to escape them using `\`. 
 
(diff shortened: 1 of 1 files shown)


quietvoid:
Maybe the credentials variables example can be updated to use single quotes?
That would avoid the need for escaping.

```
ZWIFT_PASSWORD='password'
```

gallais:
Except if the password also contains single quotes. :grin:

quietvoid:
Right, but it would be specific to the single quote case, and would still be better than the current example that isn't using quotes at all.

netbrain:
Doesn't this go without saying when things are stored in env vars?

gallais:
I guess the question is whether we are expecting all users to be devs.

netbrain:
> I guess the question is whether we are expecting all users to be devs.

Fair enough

Issue #30, opened 2023-10-11
Use with zoffline server


oldnapalm:
Hello.

Someone asked how they could use this image with a zoffline server, I found a way and posted the instructions in my fork in case someone else is interested https://github.com/oldnapalm/zwift#how-can-i-modify-the-image-to-use-with-zoffline and also added a script to launch Zwift using zoffline https://github.com/oldnapalm/zwift/blob/master/zoffline.sh

I have very little experience with Docker, there's probably a better way to do it.

PR #34, opened 2023-10-16
Add instructions to use with zoffline

oldnapalm:
I understand if you don't want to merge this because it's unofficial, but just in case...

#30


netbrain:
Yeah, I don't feel like adding official support for zoffline, not for the time being at least.

But how about you add this under show and tell instead?

https://github.com/netbrain/zwift/discussions/categories/show-and-tell

In any case I feel like this warrants its own Docker image and has its own GitHub actions to build this image. Probably tagged as a variant like netbrain/Zwift:offline-1.50.0

I'll reconsider it if you add this automation as well 😅🫣

PR #50, opened 2023-12-14
added winetricks zwift verb

netbrain:


quietvoid:
I'm not sure if there's an advantage of the verb file instead of using `/SP- /VERYSILENT /SUPPRESSMSGBOXES /NORESTART /NOCANCEL`.  
Now the installer does show up even though it's handled automatically.

On my side it also didn't close after setup finished, I'm not exactly sure why.

netbrain:
My thinking was that it would be easier to see what's going on when things
fail and know exactly where in the installation process it fails.

And when it does it might be beneficial to allow autohotkey to do extra
intermediary steps to prevent failure.

You can also run complete interactive installation by omitting the -q
argument to winetricks.

Issue #54, opened 2024-01-01
Zwift doesn't start with 1.55.0


gaeljw:
Hello,

Since this morning, I'm facing an issue with the latest image (digest ac9eb7f0a269, v1.55.0).

Zwift doesn't start and instead I get two error popups from Wine:

> Wine
> The Wine configuration in /home/user/.wine is being updated, please wait...

> rundll32.exe - This applicat...
> This application could not be started.
> Do you want to view information about this issue?

The last image working fine for me is digest ad3bca953cb1 v.1.54.0 from a few days ago. When pinning the image to this one, everything works fine.

I didn't have time to troubleshoot more, here are the container logs:
```
+ ZWIFT_HOME='/home/user/.wine/drive_c/Program Files (x86)/Zwift'
+ mkdir -p '/home/user/.wine/drive_c/Program Files (x86)/Zwift'
+ cd '/home/user/.wine/drive_c/Program Files (x86)/Zwift'
+ '[' '' = update ']'
++ ls -A .
+ '[' '!' 'ANT_DLL.dll
APR2007_xinput_x64.cab
APR2007_xinput_x86.cab
BleWin10Lib.dll
BleWin10Lib_V2.dll
CloseLauncher.exe
DSETUP.dll
DXSETUP.exe
Feb2010_X3DAudio_x64.cab
Feb2010_X3DAudio_x86.cab
Jun2010_XAudio_x64.cab
Jun2010_XAudio_x86.cab
Launcher_ver_cur.xml
Patcher.dll
VC_redist.x64.exe
WebView2Loader.dll
Webview2Setup.exe
Windows ANT Dongle Driver
ZwiftApp.exe
ZwiftHelper.exe
ZwiftLauncher.exe
ZwiftLauncher.exe.config
ZwiftWindowsCrashHandler.exe
Zwift_1.0.123458_bd4431b2_manifest.xml
Zwift_ver_cur.xml
Zwift_ver_cur_filename.txt
assets
bink2w64.dll
data
dotnet-v4.7.2-installer.exe
dsetup32.dll
dxupdate.cab
starting zwift...
unins000.dat
unins000.exe' ']'
+ echo 'starting zwift...'
+ wine64 start ZwiftLauncher.exe SilentLaunch
[...]

PR #63, opened 2024-01-17
additional package source for newer mesa-libGL

fandan:
Resolve glxinfo errors on recent Fedora host version (39) in combination with Radeon GPU

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ ENV WINEDEBUG=${WINEDEBUG:-fixme-all}
 
 RUN dpkg --add-architecture i386 
 
-RUN echo "deb http://ftp.de.debian.org/debian sid main" >> /etc/apt/sources.list
+RUN echo "deb http://ftp.debian.org/debian sid main" >> /etc/apt/sources.list
 
 RUN \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -21,6 +21,8 @@ ENV WINEDEBUG=${WINEDEBUG:-fixme-all}
 
 RUN dpkg --add-architecture i386 
 
+RUN echo "deb http://ftp.de.debian.org/debian sid main" >> /etc/apt/sources.list
```
netbrain: Maybe use the official mirror? 🤷

netbrain:
```suggestion
RUN echo "deb http://ftp.debian.org/debian sid main" >> /etc/apt/sources.list
```

netbrain (changes requested):

netbrain:
Tested it locally, seems to work fine. Releasing it to the wild to see what happens.

gaeljw:
Worked fine for me, thanks folks!

(The issue described in comments of https://github.com/netbrain/zwift/issues/54 is solved :+1: )

PR #65, opened 2024-01-19
Fix suite bullseye to bookworm

fandan:
Fix Suite in debian source list from previous bullseye to current used bookworm


netbrain:
https://github.com/netbrain/zwift/actions/runs/7581312588

PR #81, opened 2024-04-03
trying to fix latest version 1.61

netbrain:
refactored to use alpine image and wine 9.0


quietvoid:
Going back to bookworm packages would downgrade drivers for some users who needed them for newer GPUs.
It might be better to just use the Debian unstable Docker image.

netbrain:
you are probably right.

netbrain:
Merging, let's see if this creates a shitshow of issues the coming days. :hot_face:

cmuench:
@netbrain I just updated and it seems to work well on Fedora 39 KDE with Wayland and Nvidia GPU.
Before it was only working with X on my machine without any issue.
So it's a improvement. Thanks for the update.

cmuench:
@netbrain It's working with nouveau nvidia driver but not with vulkan.
I just saw that the new image is now based on alpine and the vulkan related packages are not installed anymore.
IMHO this packages are potential candidates `vulkan-tools, vulkan-loader, krb5`.

quietvoid:
Why do you need Vulkan exactly? Zwift is OpenGL.

cmuench:
> Why do you need Vulkan exactly? Zwift is OpenGL.

Good questions.
In my container log I got this messages:

```
0190:err:ole:CoGetContextToken apartment not initialised
0178:err:combase:RoGetActivationFactory Failed to find library for L"Windows.Foundation.Diagnostics.AsyncCausalityTracer"
01b0:err:vulkan:wine_vk_init Failed to load libvulkan.so.1.
```

cmuench:
I see tht the Zwift version in the container is 1.62.0.
Maybe a change between 1.60.0 and 1.62.0?
Version 1.60.0 still runs fine.

netbrain:
Yeah probably did not get all the required libraries for all uses of this image. Could you try to add the missing libraries? 

You can do this by changing the docker container directly.

- run the Zwift container with --entrypoint=bash and -it flags. 
- apk add <missing library pkg>
- re run /bin/entrypoint

If these succeeds, please add the missing libraries to the Dockerfile 🤞👍

cmuench:
@netbrain 
Just created an own docker image with the libs included but it's not running.

I just investigated the issue a but. Could be that the nvidia driver has issues to run with musl libc which is used by alpine instead of libc.

cmuench:
@netbrain interesting behaviour ... If I install the libs manually in the container as proposed by you, the error messages are then gone but the game runs still slow not using the nvidia driver.

netbrain:
> @netbrain interesting behaviour ... If I install the libs manually in the container as proposed by you, the error messages are then gone but the game runs still slow not using the nvidia driver.

Crap. 🫣

cmuench:
> > @netbrain interesting behaviour ... If I install the libs manually in the container as proposed by you, the error messages are then gone but the game runs still slow not using the nvidia driver.
> 
> Crap. 🫣

Crap is all the stuff we have to deal with Nvidia cards :-)
I need the nvidia card only for all the CUDA AI stuff. 
With AMD cards I had nearly no issues.

cmuench:
For reference: https://github.com/NVIDIA/nvidia-installer/issues/10

netbrain:
Guess we'll need a separate solution for nvidia in that case. Anyone have any recommendations for container os to use that is lightweight and easy to use when it comes to installing wine , gfx drivers etc?

Might just revert back to debian or use ubuntu or even arch.  Or maybe we should provide multiple zwift images, alpine/debian/+++ ? :thinking:

netbrain:
@cmuench you mind testing out #84 and see if this works on your nvidia card? :crossed_fingers:

PR #86, opened 2024-04-09
Fixed issue with device flag for podman and nvidia

netbrain:


netbrain:
@dmccombs you mind testing this?

PR #90, opened 2024-04-11
[ readme ] escaping now breaks config-based login

gallais:


gallais:
```
+ source /home/$USER/.config/zwift/config
++ ZWIFT_USERNAME=USER
++ ZWIFT_PASSWORD=PASS_WITH_A_SEMICOLON
/home/$USER/.config/zwift/config: line 2: PASS_AFTER_SEMICOLON: command not found
```

Hmm. Auto-login works without escaping the semicolon character (presumably because of
`ZWIFT_CONFIG_FLAG="--env-file $HOME/.config/zwift/config"`) but I do get noise on the
command line because of the call to `source`. :thinking:

netbrain (approved):

netbrain:
Lol. Probably need to escape the special character before source.

PR #91, opened 2024-04-13
Add pulse permission fixing, network mode setting & container name

hobeone:
Probably better to split these up in a few pull requests but was hoping to get initial thoughts on the various changes.  Happy to break these apart and also add updates to the readme if you're okay with the changes.

On the Pulse permissions I was hoping to allow something like what some of the linuxserver.io images allow where you can set the UID/GID to use as part of the runtime environment but that is beyond my docker skills at the moment.

I needed to set the network mode to host to get Zwift to find my WiFi enabled kicker:
https://support.wahoofitness.com/hc/en-us/articles/9211851310738-Using-KICKR-KICKR-BIKE-Wi-Fi

### DIFF
--- a/zwift.sh
+++ b/zwift.sh
@@ -33,6 +33,8 @@ else
     VGA_DEVICE_FLAG="--device /dev/dri:/dev/dri"
 fi
 
+CONTAINER_NETWORKING="bridge"
+
 ### OVERRIDE CONFIGURATION FROM FILE ###
 
 # Check for other zwift configuration, sourced here and passed on to container aswell
@@ -71,11 +73,24 @@ fi
 
 ### START ###
 
+# Give write access to pulse if UID doesn't match UID inside the container.
+# Quick fix, but there are probably better ways of doing this where you
+# dynamically set the UID of the user running zwift at runtime.  The
+# --userns=keep-id argument to podman gets around this by mapping UIDs at
+# runtime do this isn't needed when using podman
+#
+if [[ "$CONTAINER_TOOL" == "docker" && $UID != 1000 ]]
+then
+  chmod 755 "/run/user/$UID/pulse"
+fi
+
 # Start the zwift container
 CONTAINER=$($CONTAINER_TOOL run \
     -d \
     --rm \
     --privileged \
+    --network $CONTAINER_NETWORKING \
+    --name zwift-$USER \
     -e DISPLAY=$DISPLAY \
     -v /tmp/.X11-unix:/tmp/.X11-unix \
     -v /run/user/$UID/pulse:/run/user/1000/pulse \


zwift.sh
```diff
@@ -71,11 +73,24 @@ fi
...
+# Quick fix, but there are probably better ways of doing this where you
+# dynamically set the UID of the user running zwift at runtime.  The
+# --userns=keep-id argument to podman gets around this by mapping UIDs at
+# runtime do this isn't needed when using podman
+#
+if [[ "$CONTAINER_TOOL" == "docker" && $UID != 1000 ]]
+then
+  chmod 755 "/run/user/$UID/pulse"
+fi
+
```
netbrain, line 86: Definitely think this should be it's own pull request. And also we should probably just check for read access, and if  read access is not granted, then we can exit with a warning that the pulse socket is currently not readable by the container.

zwift.sh
```diff
@@ -33,6 +33,8 @@ else
     VGA_DEVICE_FLAG="--device /dev/dri:/dev/dri"
 fi
 
+CONTAINER_NETWORKING="bridge"
```
netbrain, line 36: Should probably use bridge as default network as you state, but should also be overridable like, and maybe rename to NETWORKING?

NETWORKING=${NETWORKING:-bridge}

This will keep the arguments simple. Please also add the documentation to README of the newly added NETWORKING config option.

zwift.sh
```diff
@@ -71,11 +73,24 @@ fi
...
+  chmod 755 "/run/user/$UID/pulse"
+fi
+
 # Start the zwift container
 CONTAINER=$($CONTAINER_TOOL run \
     -d \
     --rm \
     --privileged \
+    --network $CONTAINER_NETWORKING \
+    --name zwift-$USER \
```
netbrain, line 93: This seems fine :+1: 

Can't think of a reason why anyone would need anything else in regards to naming. But in the case that people would want to start two zwift containers under the same $USER then maybe this could be provided using a config option like the NETWORKING parameter aswell. Ill leave that up to you.

zwift.sh
netbrain: As for how we can implement dynamic user and group, this is probably doable, but i think it would require that the entrypoint script actually changes the group and user of the /home/user directory as it's first directive upon launching the container. Where the entrypoint script simply would have to check if UID is equal to /home/user and if not, then `usermod -u xxx foo; groupmod -g xxx foo; chown -r x:y  /home/user`

zwift.sh
netbrain: This could be useful if someone needs host networking or their own custom networking

hobeone:
Thanks for the comments.  I'm going to close this pull request and rework into multiple smaller ones.

PR #92, opened 2024-04-19
Add ability to set container networking mode

hobeone:
The main use case would be to set it to "host" networking mode which allows Zwift to see Wifi enabled trainers on the same network.

See this link for the Wahoo example:
https://support.wahoofitness.com/hc/en-us/articles/9211851310738-Using-KICKR-KICKR-BIKE-Wi-Fi


netbrain:
Lgtm

Issue #95, opened 2024-04-24
Permission denied for groupmod with podman


fandan:
When I start the container I get the following error message:

```
++ id -u user
+ USER_UID=1000
++ id -g user
+ USER_GID=1000
+ '[' -n 1000 ']'
+ '[' 1000 -eq 1000 ']'
+ USER_UID=1000
+ '[' -n 1000 ']'
+ '[' 1000 -eq 1000 ']'
+ USER_GID=1000
+ usermod -o -u 1000 user
usermod: no changes
+ groupmod -o -g 1000 user
groupmod: Permission denied.
groupmod: cannot lock /etc/group; try again later.
```

I suspect this comes from using gosu in entrypoint.sh and removing the sudo param. gosu doesn't seem to be supported by podman. Few days ago zwift runs without problems.

netbrain:
@hobeone ☝️

hobeone:
Will take a look but I suspect that it could be related to the line zwift.sh:

```    $([ "$CONTAINER_TOOL" = "podman" ] && echo '--userns=keep-id') ```

I don't think that's necessary anymore after the changes in #94.  I've never used podman but looking at the [docs](https://docs.podman.io/en/v4.4/markdown/options/userns.container.html) I think either deleting it or setting it to the empty string might work.

My guess is that when you set `--userns=keep-id` entrypoint isn't running as root in the container as it does with docker.

@fandan could you give that a shot and see if it gets it working?

nibbles-bytes:
> Will take a look but I suspect that it could be related to the line zwift.sh:
> 
> ` $([ "$CONTAINER_TOOL" = "podman" ] && echo '--userns=keep-id')`
> 
> I don't think that's necessary anymore after the changes in #94. I've never used podman but looking at the [docs](https://docs.podman.io/en/v4.4/markdown/options/userns.container.html) I think either deleting it or setting it to the empty string might work.
> 
> My guess is that when you set `--userns=keep-id` entrypoint isn't running as root in the container as it does with docker.
> 
> @fandan could you give that a shot and see if it gets it working?

I just tried removing that line in my podman setup and I was able to boot Zwift successfully.

fandan:
@nibbles-bytes , @hobeone: Removing `--userns=keep-id` solved the problem. Now zwift is starting again. Thanks.

PR #98, opened 2024-04-25
Removed podman userns flag

netbrain:
This seems to cause issues for podman users after #94,  as described in #95


netbrain:
@fandan  @hobeone  @nibbles-bytes :point_up: 

can you guys confirm that this actually solves the issue?

netbrain:
Container image should be updated after github action workflow has been run. :crossed_fingers:

Issue #100, opened 2024-04-25
pulseaudio doesn't work after running zwift


perrin4869:
**Describe the bug**
Due to #94, after running zwift, pulseaudio will fail with a message:

```
$ pamixer
Failed to create secure directory (/run/user/1000/pulse): Permission denied
Connection error
```

Looking at the directory, I find:

```
 $ ls -lah /run/user/1000
total 4.0K
drwx------ 17 perrin4869 users  500 Apr 25 22:25 .
drwxr-xr-x  3 root       root    60 Apr 25 12:48 ..
drwx------  2 perrin4869 users   60 Apr 25 12:48 at-spi
drwx------  6 perrin4869 users  120 Apr 25 17:37 containers
drwx------  3 perrin4869 users   60 Apr 25 12:48 dbus-1
drwx------  2 perrin4869 users   60 Apr 25 12:48 dconf
dr-x------  2 perrin4869 users    0 Jan  1  1970 doc
drwx------  2 perrin4869 users  140 Apr 25 12:50 gnupg
dr-x------  2 perrin4869 users    0 Apr 25 12:48 gvfs
drwx------  2 perrin4869 users   40 Apr 25 17:17 gvfsd
drwx------  2 perrin4869 users  100 Apr 25 12:48 i3
drwx------  2 perrin4869 root    60 Apr 25 22:25 keyring
drwx-----T  3 perrin4869 users   60 Apr 25 17:37 libpod
drwx--x--x  2 perrin4869 users   60 Apr 25 19:26 netns
srwxr-xr-x  1 perrin4869 users    0 Apr 25 17:18 nvim.17872.0
srwxr-xr-x  1 perrin4869 users    0 Apr 25 14:14 nvim.24940.0
srwxr-xr-x  1 perrin4869 users    0 Apr 25 12:50 nvim.6123.0
srwxr-xr-x  1 perrin4869 users    0 Apr 25 12:48 pipewire-0
srwxr-xr-x  1 perrin4869 users    0 Apr 25 12:48 pipewire-0-manager
-rw-r-----  1 perrin4869 users    0 Apr 25 12:48 pipewire-0-manager.lock
-rw-r-----  1 perrin4869 users    0 Apr 25 12:48 pipewire-0.lock
drwx------  2     100999 100099  80 Apr 25 12:48 pulse
-rw-------  1 perrin4869 users    5 Apr 25 22:24 rofi.pid
[...]

PR #102, opened 2024-04-29
fix: podman compatibility

perrin4869:
Fixes https://github.com/netbrain/zwift/issues/100
The workings of user namespaces in Linux elude me, so I think it's still important to get to the root causes, but at least this will restore `podman` compatibility as it was before

### DIFF
--- a/zwift.sh
+++ b/zwift.sh
@@ -26,7 +26,7 @@ then
     if [[ $CONTAINER_TOOL == "podman" ]]
     then
         VGA_DEVICE_FLAG="--device=nvidia.com/gpu=all"
-    else 
+    else
         VGA_DEVICE_FLAG="--gpus all"
     fi
 else
@@ -76,7 +76,6 @@ fi
 
 ### START ###
 
-# Start the zwift container
 CONTAINER=$($CONTAINER_TOOL run \
     -d \
     --rm \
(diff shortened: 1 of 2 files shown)


zwift.sh
```diff
@@ -76,14 +76,15 @@ fi
...
-# Start the zwift container
 CONTAINER=$($CONTAINER_TOOL run \
     -d \
     --rm \
     --privileged \
     --network $NETWORKING \
     --name zwift-$USER \
     -e DISPLAY=$DISPLAY \
+    $([ "$CONTAINER_TOOL" = "podman" ] && echo '--userns=keep-id') \
+    $([ "$CONTAINER_TOOL" = "podman" ] && echo '--entrypoint /bin/setup_and_run_zwift') \
```
netbrain, line 87: For what reason is the entrypoint overriden?

netbrain (commented):
Thank you for your contribution, looks good, and ill take your word for it that this restores podman behavior. However one question before i approve the PR.

zwift.sh
netbrain: Ah, i see the reason now, by overriding you skip the whole change of UID and GID of the entrypoint. I guess this is ok. But would you mind adding a section in the readme that describes that ZWIFT_UID and GID is only applicable to docker and not podman?

perrin4869: right, let me add that!

netbrain (approved):

netbrain (approved):

PR #105, opened 2024-04-29
feat: inhibit screensaver

perrin4869:
since running zwift doesn't require any keyboard/mouse input, the screensaver/screen lock, etc may be triggered in the middle.
afaik, the standard for inhibiting the screensaver is to periodically call the dbus service `org.freedesktop.ScreenSaver` with the `Inhibit` method, for example, with [`xscreensaver`](https://github.com/Zygo/xscreensaver/blob/0c43268adc2e7a932ca5427db7b18d37ef53f7ad/driver/xscreensaver-systemd.c#L41)
This is what Firefox does.
The easiest way to achieve this, is to use `xdg-screensaver reset` from the `xdg-utils` package.
I've been running this separately in another terminal each time I start `zwift`, but I think it's a good idea to add support for it for anyone who's running a dbus server.

Edit: lutris may have done something similar: https://github.com/lutris/lutris/issues/2424#issuecomment-609976923

### DIFF
--- a/setup_and_run_zwift.sh
+++ b/setup_and_run_zwift.sh
@@ -66,7 +66,7 @@ then
 
     # install dotnet48 for zwift
     winetricks -q dotnet48
-        
+
     # install webview 2
     wget -O webview2-setup.exe https://go.microsoft.com/fwlink/p/?LinkId=2124703
     wine webview2-setup.exe /silent /install
@@ -79,7 +79,7 @@ then
     # update game through zwift launcher
     wait_for_zwift_game_update
     wineserver -k
-    
+
     # cleanup
     rm "$ZWIFT_HOME/ZwiftSetup.exe"
     rm "$ZWIFT_HOME/webview2-setup.exe"
@@ -115,6 +115,8 @@ do
     echo "Waiting for zwift to start ..."
     sleep 1
 done
+
+[[ -n "${DBUS_SESSION_BUS_ADDRESS}" ]] && watch -n 30 xdg-screensaver reset &
 
 echo "Killing uneccesary applications"
 pkill ZwiftLauncher
--- a/zwift.sh
+++ b/zwift.sh
@@ -76,12 +76,12 @@ fi
 
 if [[ -n "$DBUS_SESSION_BUS_ADDRESS" ]]
 then
-    REGEX='^unix:path=([^,]+)'
-    [[ $DBUS_SESSION_BUS_ADDRESS =~ $REGEX ]]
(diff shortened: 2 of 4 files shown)


zwift.sh
```diff
@@ -74,6 +74,17 @@ then
     $CONTAINER_TOOL pull $IMAGE:$VERSION
 fi
 
+if [[ -n "$DBUS_SESSION_BUS_ADDRESS" ]]
+then
+    REGEX='^unix:path=([^,]+)'
```
netbrain: Maybe a better naming convention here?

setup_and_run_zwift.sh
```diff
@@ -116,6 +116,8 @@ do
     sleep 1
 done
 
+[[ -n "${DBUS_SESSION_BUS_ADDRESS}" ]] && watch -n 30 xdg-screensaver reset &
```
netbrain, line 119: Will this command prevent the container from exiting when Zwift is closed? I'm guessing no, but have to ask.

netbrain (commented):
Looks good, but maybe add a line about this in the README.

setup_and_run_zwift.sh
perrin4869: So I had a similar concern, but it did exit the container, at least in `podman`, when I tested it.
I think the container exists as soon as the `entrypoint` finishes executing, regardless of running processes.

zwift.sh
perrin4869: You mean the `DBUS_SESSION_BUS_ADDRESS` or `REGEX`?
`DBUS_SESSION_BUS_ADDRESS` is exposed by the dbus daemon, for example, by running

```
$ dbus-launch --exit-with-x11                                                                                                                                            
DBUS_SESSION_BUS_ADDRESS=unix:path=/tmp/dbus-GTNRHNKqos,guid=c5d34141f3a711e5c1b93770662fc7a8
DBUS_SESSION_BUS_PID=6066
```

You will probably have these two setup in your own Linux environment.

perrin4869:
hm... well, as it is, it is completely transparent from the user perspective, so I'm not sure what I would mention in the readme, other than to add a paragraph introducing this feature?

zwift.sh
netbrain: Regex

netbrain:
Maybe just mention that every 30 seconds x screensaver is inhibited on the host machine if the env variable is defined.

zwift.sh
perrin4869: right, I made that segment a bit cleaner :)

netbrain (approved):

Issue #96, opened 2024-04-24
Refactor script to separate out starting and updating zwift


hobeone:
As described in pull request #94 it would be good to split out the code that starts Zwift vs the code that updates it.  Filling a bug to understand the background and requirements.

@netbrain what did you want to get out of this refactor?

netbrain:
The entrypoint script has two responsibilities today. 

1. Update game files, which it does on the first run, or when update argument is present.
2. Start Zwift.

Would be good to split this into a update.sh and a start.sh. this would further modularize this solution and make it possible for reuse in other use cases.

hobeone:
Working through this now.

Is there a reason that the script exits after installing?  I would expect it to install Zwift if it isn't already there and then start it.

netbrain:
It's probably due to how the github actions are set up and how we build the zwift container (which you know contains all game files)

So first we run the "empty  zwift image" which runs with the update flag and simply downloads the latest zwift files. Then we commit's that container instance to a new image i.e zwift:latest and zwift:<version>.

PR #108, opened 2024-05-06
Refactor updating/installing/running Zwift into their own scripts.

hobeone:
First pass at addressing #96

### DIFF
--- a/zwift_common.sh
+++ b/zwift_common.sh
@@ -1,1 +1,0 @@
-ZWIFT_HOME="$HOME/.wine/drive_c/Program Files (x86)/Zwift"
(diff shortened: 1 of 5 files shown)


zwift_common.sh
netbrain, line 1: this seems a bit overengineered when it only holds a single environment var and sourced into two other files. Maybe duplication is the better answer here for the time being?

netbrain: These script files also need to be added to the zwift_push yaml for github actions.

netbrain (commented):
Mostly looks good, need to be added to the github action file for script changes tho.

hobeone:
Added to the github action file. 

Also fixed an issue in entrypoint where we need to find the home directory of "user" and can't just use $HOME because it's being run as root.

I've run the docker image in both normal and "update" mode and it seems to be doing the right thing.

hobeone:
Hmm.  Is there a skew between the Dockerfile that the workflow uses and head?  Does the new Dockerfile need to be pushed somehow so it get's pulled when the workflow does `docker pull netbrain/zwift:latest` ?

PR #113, opened 2024-05-07
Various Fixes for Podman/ Docker

sHedC:
From the latest have updated to fix the following, mainly for podman but tested with both docker and podman on xwayland, x11, wayland in both KDE (Manjaro) and Gnome (Fedora).

- podman/ docker now detected in entrypoint and script split to start correctly.
- properly supports xwayland using xAuthority
- fixed permissions when setting up new zwift user volume for podman
- added ability to specify user and group id in podman
- can locally build and run in podman using bin/test-build.sh (also works for docker)
- fixed entrypoint for docker initial build when using same userid as host
- fixed passing of arguments from test-build.sh and zwift scripts

Wayland i tried to fix, it starts wine notepad however zwift complains about screen size and doesn't work correctly.


run_zwift.sh
```diff
@@ -16,6 +16,7 @@ if [[ ! -z "$WINE_EXPERIMENTAL_WAYLAND" ]];
 then
     echo "enabling wayland support in wine 9.0"
     wine reg.exe add HKCU\\Software\\Wine\\Drivers /v Graphics /d x11,wayland
+    DISPLAY=
```
netbrain, line 19: nice catch

zwift.sh
```diff
@@ -20,25 +20,30 @@ then
...
+
+    -e DISPLAY=$DISPLAY
+    -e WINE_EXPERIMENTAL_WAYLAND=$WINE_EXPERIMENTAL_WAYLAND
+    -e ZWIFT_UID=$ZWIFT_UID
+    -e ZWIFT_GID=$ZWIFT_GID
+    -e XAUTHORITY=$XAUTHORITY
+
+    -v /tmp/.X11-unix:/tmp/.X11-unix
+    -v /run/user/$UID:/run/user/$ZWIFT_UID
+    -v zwift-$USER:/home/user/.wine/drive_c/users/user/Documents/ZwiftX
```
netbrain: ZwiftX?

zwift.sh
```diff
@@ -64,7 +70,12 @@ then
     if [ "$REMOTE_SUM" = "$THIS_SUM" ]; then
         echo "You are running latest zwift.sh 👏"
     else
-        echo "You are not running the latest zwift.sh 😭, please update!"
+        RED='\033[0;31m'
+        NC='\033[0m'
+        BOLD='\033[1m'
+        UNDERLINE='\033[4m'
+        echo -e "${RED}${BOLD}${UNDERLINE}You are not running the latest zwift.sh 😭, please update!${NC}"
+        sleep 5
```
netbrain, line 78: stylish :)

netbrain (commented):
Your bash skills are a notch above mine :+1: 

lgtm,. however one question.

netbrain:
@hobeone do you have any inputs on this?

zwift.sh
sHedC: Oops misstype I'll check it when I am home later

entrypoint.sh
```diff
@@ -2,41 +2,57 @@
...
+    groupmod -o -g ${USER_GID} user
+    chown -R ${USER_UID}:${USER_GID} /home/user
 
-mkdir -p /run/user/${USER_UID}
-chown -R user:user /run/user/${USER_UID}
-sed -i "s/1000/${USER_UID}/g" /etc/pulse/client.conf
+    if [ ! -d "/run/user/${USER_ID}" ]; then
+        mkdir -p /run/user/${USER_UID}
+        chown -R user:user /run/user/${USER_UID}
+        sed -i "s/1000/${USER_UID}/g" /etc/pulse/client.conf
```
hobeone, line 56: Should the sed command run always (i.e. not in the if statement?

entrypoint.sh
```diff
@@ -2,41 +2,57 @@
 set -e
 set -x
 
-# This script runs as the root user in Docker so need to do this to find the
-# home directory of the "user" user.
-ZWIFT_USER_HOME=$( getent passwd "user" | cut -d: -f6 )
-ZWIFT_HOME="$ZWIFT_USER_HOME/.wine/drive_c/Program Files (x86)/Zwift"
+# Check what container we are in:
+if [[ -z $(cat /run/.containerenv | grep "podman") ]]; then
```
hobeone: Would it be better to check for existence of /.dockerenv since your basically testing to see if you're running in docker?

hobeone:
Only other input is Thank You for doing all of this!

entrypoint.sh
sHedC: No idea :)  what is this rare case for?  couldn't test it, I will move it ousdie of the if.

entrypoint.sh
sHedC: podman is in .containerenv, might be better to pass in an env.  I will check.

sHedC:
> Your bash skills are a notch above mine 👍
> 
> lgtm,. however one question.

I have been coding for over 35 years, I actually was trying not to change too much :)
I am more than happy to help with Podman integration and Wayland as this is what I use for my other developments.

entrypoint.sh
sHedC: @hobeone - If you are setting the user to be a different ID inside the docker wouldn't userns bet an approach, this is the way of doing it in podman.  In the doc it says about using a specific case for setting a different ID, wondered what it was so I can test.

sHedC:
Fixed the comments above, I have updated bin/test_build.sh to trap for errors and clean up after (annoying if you have to manually delete the container on failure).

I didn't update the bin/build.sh but the test one should work for both Podman and Docker, if you are happy I will move it into build.sh and update the readme.

I am just running docker tests now.

entrypoint.sh
hobeone: Perhaps.  I am by no means a docker expert.  I was looking at how some other docker containers I use approach this and settled on this way.  For me I kicked off these changes so that the UID/GID of the user in the docker container matched my own UID/GID so that sound worked.

If there is a better way of doing this in docker I'd be happy to learn how and maybe we change change it in a follow up change?

sHedC: Not sure, there is a possiblity of using re-mapping I use it for containers in proxmox but whether it will work don't know, I was going to try later with podman, I know podman better than docker as I use it for dev work, but use lxc containers quite a lot.

sHedC:
Ok docker is working but when I try using a different ZWIFT_UID/GID I get an error later when starting, this might just be because I am not using it right as not sure why it would be used.

The entrypoint is setting all the permissions correctly and setting up the right mapping.  Can you seee if that part works?

hobeone:
What error are you seeing?

sHedC:
> What error are you seeing?

image

netbrain:
Looks like an error from the runfromprocess-rs binary. @quietvoid

quietvoid:
The process PID is probably invalid.

sHedC:
This is only when I put ZWIFT_UID and ZWIFT_GID different to my local id of 1000.  Without setting those it works fine.

sHedC:
ok got it, because I mount /run/user/??? directly it is not setting the ownership correctly. Problem is /run/user//?? is neede for wayland, need to re-adjust.

sHedC:
Have changed it back a bit, so basically to use Wayland or XWayland we need to map /run/user/xxxx, however doing this means you can not change ownership of /run/user/xxxx to the UID/ GID provided in ZWIFT_UID and GID.

So in short at the moment if you use ZWIFT_UID / GID you have to use X11, but wayland works for default.

Will look into this more but fixed to Podman/ Docker works with defefault in Wayland and X11 but adding ZWIFT_UID/GID requires using X11

netbrain (approved):

sHedC:
Found a couple of minor issues with podman savng documents and sound since fixing the ZWIFT_UID issue.

sHedC:
> Nice, is it all done now or are there any remaining issues?

Just testing and one more I notice the update script should e reversed, "UPDATE" should be checked after install is checked.

sHedC:
Fixed Tested with ZWIFT_UID and docker (podman fails for some weird reson but nearly works)
Fix Audio on Docker
Build now supports docker and podman
Fixed Podman permissions

Tested everything I can

bin/build-image.sh
```diff
@@ -1,6 +1,45 @@
...
+fi
+
+# Update information based on Container Tool
+if [[ "$CONTAINER_TOOL" == "podman" ]]; then
+    BUILD_NAME="zwift"
+    IMAGE="localhost/zwift"
+else
+        BUILD_NAME="netbrain/zwift"
+        IMAGE="netbrain/zwift"
+fi
```
netbrain, line 27: What is the reasoning behind the different names? i.e why when podman you have localhost/zwift instead of netbrain/zwift? relic from testing?

update_zwift.sh
```diff
@@ -87,3 +82,12 @@ then
     rm -rf "$HOME/.cache/wine*"
     exit 0
```
netbrain, line 83: Not for future, this could have been extracted into it's own install_zwift.sh

zwift.sh
```diff
@@ -151,7 +151,12 @@ then
     )
 fi
 
-CONTAINER=$($CONTAINER_TOOL run ${GENERAL_FLAGS[@]} \
+# If we are using a different user enable xhost.
+if [ $ZWIFT_UID -ne 1000 ]; then
+    xhost +
+fi
```
netbrain: Any reason why you need to enable xhost? (should also probably be +local?) both above an below the actual runneth of the container?
 
 Why doesn't the below suffice?

netbrain (commented):
A few questions & clarifications.

zwift.sh
sHedC: I will try again but it seemed not to work for the specific case of changing the UID.

sHedC:
Weirdly Podman now supports using ZWIFT_UID under X11 not sure how that happened.

bin/build-image.sh
netbrain: Think you forgot to answer this

netbrain: @sHedC :point_up:

sHedC:
I this all good now anything else you notice?

netbrain:
> What is the reasoning behind the different names? i.e why when podman you have localhost/zwift instead of netbrain/zwift? relic from testing?

sHedC:
> > What is the reasoning behind the different names? i.e why when podman you have localhost/zwift instead of netbrain/zwift? relic from testing?
> 
> 

Think you missed my answer or it's not commenting properly from my phone.

Podman always prefixes with localhost automatically unless you build against a real domain. So wasn't really a choice to use localhost for podman local build.

netbrain:
Thanks for your contribution, The community appreciates it!

netbrain:
Just got around to test this myself now, seems to work well. :clap:  Have released this as a new image built from scratch. So for the next upcoming releases we'll see if the updating routine works aswell.

sHedC:
> Just got around to test this myself now, seems to work well. 👏 Have released this as a new image built from scratch. So for the next upcoming releases we'll see if the updating routine works aswell.

Perfect glad I can help i am more than willing to help in the future I use this a few times a week and it was one of the last things I used windows for, really appreciate the effort you have all putin to make this work so happy I can assist.

netbrain:
Happy to hear! 🤗

PR #117, opened 2024-05-13
chown step regarding ZWIFT_UID and ZWIFT_GID is run regardless  of equality within container in docker.

### LINKED ISSUE #116: chown step regarding ZWIFT_UID and ZWIFT_GID is run regardless  of equality within container in docker.
**Describe the bug**
Zwift startup hangs on chown of /home/user within container

**To Reproduce**
1. Use docker
2. Execute `ZWIFT_UID=1000 ZWIFT_GID=1000 zwift.sh`
3. Execute `docker logs zwift-$USERNAME`
4. See that the last log line within a few seconds is the chown instruction

```
+ [[ -f /.dockerenv ]]
+ CONTAINER=docker
+ [[ docker == \d\o\c\k\e\r ]]
++ getent passwd user
++ cut -d: -f6
+ ZWIFT_USER_HOME=/home/user
+ ZWIFT_HOME='/home/user/.wine/drive_c/Program Files (x86)/Zwift'
+ mkdir -p '/home/user/.wine/drive_c/Program Files (x86)/Zwift'
+ cd '/home/user/.wine/drive_c/Program Files (x86)/Zwift'
++ id -u user
+ USER_UID=1000
++ id -g user
+ USER_GID=1000
+ '[' -n 1000 ']'
+ '[' 1000 -eq 1000 ']'
+ USER_UID=1000
+ '[' -n 1000 ']'
+ '[' 1000 -eq 1000 ']'
+ USER_GID=1000
+ usermod -o -u 1000 user
usermod: no changes
+ groupmod -o -g 1000 user
+ chown -R 1000:1000 /home/user
```

**Expected behavior**
If 1000 eq 1000 on both ZWIFT_UID and ZWIFT_GID it should not run the chown instruction

This is a bug because it causes zwift to take quite some time to start the ui due to this chown business. I think it only is an issue for docker however. so podman should still be fast.

sHedC:
Fixes #116

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -20,47 +20,50 @@ if [[ "$CONTAINER" == "docker" ]]; then
     # home directory of the "user" user.
     ZWIFT_USER_HOME=$( getent passwd "user" | cut -d: -f6 )
     ZWIFT_HOME="$ZWIFT_USER_HOME/.wine/drive_c/Program Files (x86)/Zwift"
-
     mkdir -p "$ZWIFT_HOME"
     cd "$ZWIFT_HOME"
 
     USER_UID=`id -u user`
     USER_GID=`id -g user`
 
-    # Test that it exists and is a number
-    if [ -n "$ZWIFT_UID" ] && [ "$ZWIFT_UID" -eq "$ZWIFT_UID" ]; then
+    # Test that it exists and is a number, and only if different to existing.
+    if [ "$ZWIFT_UID" -eq "$ZWIFT_UID" ] && [ "$USER_UID" -ne "$ZWIFT_UID" ]; then
         USER_UID=$ZWIFT_UID
+        SWITCH_IDS=1
     else
         echo "ZWIFT_UID is not set or not a number: '$ZWIFT_UID'"
     fi
-    if [ -n "$ZWIFT_GID" ] && [ "$ZWIFT_GID" -eq "$ZWIFT_GID" ]; then
+    if [ "$ZWIFT_GID" -eq "$ZWIFT_GID" ] && [ "$USER_GID" -ne "$ZWIFT_GID" ]; then
         USER_GID=$ZWIFT_GID
+        SWITCH_IDS=1
     else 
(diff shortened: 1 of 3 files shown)


entrypoint.sh
```diff
@@ -20,47 +20,50 @@ if [[ "$CONTAINER" == "docker" ]]; then
...
     mkdir -p "$ZWIFT_HOME"
     cd "$ZWIFT_HOME"
 
     USER_UID=`id -u user`
     USER_GID=`id -g user`
 
-    # Test that it exists and is a number
-    if [ -n "$ZWIFT_UID" ] && [ "$ZWIFT_UID" -eq "$ZWIFT_UID" ]; then
+    # Test that it exists and is a number, and only if different to existing.
+    if [ "$ZWIFT_UID" -eq "$ZWIFT_UID" ] && [ "$USER_UID" -ne "$ZWIFT_UID" ]; then
```
netbrain, line 30: is `[ "$ZWIFT_UID" -eq "$ZWIFT_UID" ] `really necessary? it will always result to true in the case both is an integer, and false otherwise. But the number check is also done by `[ "$USER_UID" -ne "$ZWIFT_UID" ];` is it not? 

The same applies to $ZWIFT_GID

netbrain (commented):
one question, looks good otherwise.

entrypoint.sh
sHedC: I left this in: `[ "$ZWIFT_UID" -eq "$ZWIFT_UID" ]` because it is the standard way of testing if the variable is a number, but might not be nessesary, I can try to remove and see.

sHedC: Ok not sure I can remove this I did think about it but if USER_UID = 1000 and ZWIFT_UID = TEST then it would pass as they are not equal ZWIFT_UID would evaluate to 0.

sHedC: PS I just updated the zwift.sh to do a check that the container actually started and updated the xhost to run even in wayland as it seems sometimes its needed.

PR #127, opened 2024-05-15
fix: xauthority file needs to be available in x11

perrin4869:
this restores the check that I removed yesterday, and correctly passes down the xauthority cookie to the container, I tested this and it is finally working on my system 😅
@hobeone I'm guessing you are running wayland?


netbrain (approved):

netbrain:
who needs apt-get, yum, npm, or all the other package managers when we clearly can just use containers instead :rofl: 

We should consider creating our own OS based upon a container os (i.e coreos)  and then simply have all the things in containers :laughing:

sHedC (approved):
Happy to approve also will use this going forward to re-factor the start script to be more logical in its progression.

PR #129, opened 2024-05-16
fix(xauth): add hostname to container

perrin4869:
turns out, hostname is required to match in order for xauth to work after all, sorry for the continued fixes, I think this is the last one 😅


netbrain:
No, thank you. It's wonderful to see all these changes improving the project, especially since I've been working alone for so long :)

Issue #28, opened 2023-10-05
Nix flake


netbrain:
Would be nice to have a nix flake that could build several variants of Zwift.

- [ ] A variant that would build a container as today
- [ ] A variant that would build a minimal container that uses a volume for Zwift and on first run it will simply start the Zwift download silently
- [ ] A variant that would install wine binary and install Zwift and it's dependencies and launch Zwift on host machine instead of a container
- [ ] A variant that would compile wine (statically if possible) and launch Zwift on host instead of container

Could then have a GitHub action with matrices build that would compile/install wine for different architectures, both for containers but also native builds.

netbrain:
I'm slowly learning nix, so not sure if all is possible, but it would be great if anyone with nix experience could help in this matter.

netbrain:
quietvoid/runfromprocess-rs#1

relief-melone:
I just added some edits to the flake to make zwift configurable. I think it would be great though if we could build a package that does not use a docker container as well. Might give it a go if I find the time. Question would be if it should go here though or directly to nixpkgs.
While I am starting to get the hang of NixOS I am still lacking a deep knowledge of wine. Where would I find the settings you made to wine to make zwift run this well? I tried before making zwift run in heroic and failed miserably :D

netbrain:
That's great, ill take a look at it thanks!" 

I concur, it would be great if we could leverage the nix package system to create several versions of the zwift build, just as i mentioned in the checkbox list up above. Whereas the two last points would be a host only install, and the two first, would be container builds.

The difficult part to this is that nix packages likes to lock things down, i.e no network connectivity within builds. So it makes it challenging to actually create the zwift build, but with enough vigor one should be able to work around this by i.e pre-downloading all the zwift assets beforehand.

As to your question regarding wine, the first you need to look at is the [entrypoint](https://github.com/netbrain/zwift/blob/7080d431c1828b5c961852d1a2d80be392a09d30/entrypoint.sh)

This prepares the container runtime to run zwift, and then depending if it's the first time it runs this container or it detects zwift from a previous run it will either execute [run_zwift](https://github.com/netbrain/zwift/blob/7080d431c1828b5c961852d1a2d80be392a09d30/run_zwift.sh) or [update_zwift](https://github.com/netbrain/zwift/blob/7080d431c1828b5c961852d1a2d80be392a09d30/update_zwift.sh)

[...]

PR #130, opened 2024-05-16
feat: add makefile

perrin4869:
I put together a simple makefile and added a desktop file and an icon I found online.
This would also help in packaging the app together for different distributions. I can put something together for Slackware, someone could put something together for arch linux maybe? Should help with visibility.


Makefile
```diff
@@ -0,0 +1,33 @@
...
+
+.PHONY: all install uninstall install_zwift install_icons install_desktop uninstall_zwift uninstall_icons uninstall_desktop
+
+all:
+	@echo "Call \"make install\" to install this program."
+	@echo "Call \"make uninstall\" to remove this program."
+
+install_zwift:
+	mkdir -p ${DESTDIR}${PREFIX}/bin
+	install -m 755 zwift.sh ${DESTDIR}${PREFIX}/bin/zwift
```
netbrain: Why not /usr/local/bin as in install.sh? Seems counterintuitive to place this in two different locations.

netbrain: And do we really need a makefile? could not this just be included in the install.sh we already have and refactor that to take arguments like 

1. install
2. uninstall

And just be default install desktop + icon?

Seems a bit overkill to have all these targets simlpy to remove or copy files around.

netbrain: @sHedC do you have any opinions here?

netbrain:
And to answer you comment on creating different packages, yes that would be great, and I have already created in issue around this #28 

The idea was to use nix flake in order to create different target packages and optionally different versions of the docker image. 

Like maybe someone would like to have all the zwift files downloaded (as we do today), maybe someone else simply want to have the core zwift and download the zwift files themselves on first run and save it to a volume and continuously update it locally only. Others might want to build zwift entirely from scratch. Others might want to install zwift locally on their host machine instead of going through the container layer. etc ..

all of these options could be created in a nix flake and alternatively also create packages in different distributions, (deb, rpm, etc)

Makefile
sHedC: /usr/local/bin is the right place, I looked into this before /usr/bin is for distribution managed. /user/local/bin is for administartor installed (outside of distribution) We should stick with this. This is something that has evolved since debian 9.

The make idea I have only ever used it for building and compiling from scratch after a clone so I would advise overkill for just copying and setting up a batch file to download and launch a docker.  Updating using make is also more complex than just re-copying the file.

PS in update branch 126 I was tidying up the zwift and adding an option to allow a user to say y to download the new scripts, again easier with bash files than make.

I like the idea of desktop icon/ file, but I only know gnome never done this for KDE/ XFCE etc, but again its just a simple download and copy to a specific location and is easy to do in a bash script.

sHedC:
It might be a better approach to flatpak it?  more approachable for non-technical?

Makefile
perrin4869: If a user really wants to install this package to his system using this makefile, then that's right, the standard is to install it to the `/use/local` prefix. They'd do it using `make install PREFIX=/use/local`, that's the standard way that makefiles are used 😅
Usually you don't install files into system directories by yourself though, but via a package manager, so it's not advisable to do that in the first place.

In a typical packaging scenario you'd do something like: 

```
mkdir build
make install DESTDIR=build
```

And then package it up as required by your package manager.

perrin4869: Ah right sorry, it's slightly different, right, the default should be `usr/local` and packagers should change it to `usr`

perrin4869:
Well, having a flatpak would be great for convenience, but since you need either docker or podman installed first, I'm not sure it'd be much use?

Makefile
sHedC: Only remaining comment yse I see and understand your point on build, to netbrain's comment it seems overkill for a script that just starts a docker, but if you want to full build the entire environment yourself I think this is the right way.

However in the interest of making it easy for the majority of people who are less technically minded is a single command like today better?

I am going to look to see if its possible to do everything we have done in a FlatPak, my preference as in 95% of my use case flatpak works, including sunshine for game streaming, and is easy and easier to maintain for the maintainer.  My biggest concern is the majority or work is around XOrg/ Wayland/ DBUS etc to get it working on each system and maybe Flatpak can help (no idea yet).

Ultimately it is the maintainers decision so I am going to fully support what netbrain decides.

@netbrain - Any other considerations you wish to descide I think we have:

1. Move to make to install the start script and or move to make to do a full build/ install
2. Keep current state
3. Other, e.g. have a poll.

On a final note I think it is really good to have these discussions feels more community :)

Just to add don't see any rush in making a decision on this, lets look at best approaches, I always opt for ease of use for end users this is our target otherwise what is the point :)  I implement all the time and always go for ease of use over proper approach if there is a conflict but no reason we can't do both.

sHedC:
https://flathub.org/apps/com.fightcade.Fightcade 
this uses wine.

perrin4869:
yeah I think having flatpak that you can click on and run zwift on would be pretty sick and much easier for most people (bar having automatic updates, etc), the only concern I have is that it is a huge departure from the concept of this repo 😅
It's its own separate project almost I think? since it would share little of the code available here

Makefile
perrin4869: I think that the main benefit behind a makefile over the bash script is that it provides a familiar interface for packaging, with standard parameters such as `PREFIX`, `DESTDIR` and `DATADIR`. For example, I could use it to install to my local directory by doing `make install PREFIX=~/.local` and have everything setup for me, and I could use it for packaging with the `DESTDIR` option, or use it to install it system wide by just running `sudo make install`... I kinda don't see the benefit of replicating this in bash

perrin4869: one more note, I took partly inspiration from [this makefile](https://github.com/carnager/rofi-pass/blob/master/Makefile) which is also not doing much more than installing a bash script

sHedC:
> yeah I think having flatpak that you can click on and run zwift on would be pretty sick and much easier for most people (bar having automatic updates, etc), the only concern I have is that it is a huge departure from the concept of this repo 😅 It's its own separate project almost I think? since it would share little of the code available here

Yes going to ignore it for now would prefer to focus on what is useful for this repo.  But might in future just as really fancy learing flatpak stuff.

Makefile
sHedC: I still think a make file to just install the script is overkill, as its not really its intent.  The install is usually part of a build process.

However that said I can see the benefit of make for the build/ deployment process for those people who want to, although I suspect a limited market.

perrin4869: Make is a task runner, copying files is a task... What is overkill about it? 🙂
Is there a benefit to running a bash script instead?

sHedC: ok assume it is not going to be used to build the docker, because basically the docker is built using docker file and a script built into the docker after build so 3 commands.

Also that netbrain wants to use nix flake to do the build in future.

So we have moved from a single copy/ paste command to get a script that just copies a batch file and alternatively  can also get and setup a desktop icon.

To having to clone the directory using git, change into it and run make get all files just to copy the file.

I would not bother unless I was actually building for myself, and that takes a long time.

sHedC: update would also take longer you have to pull teh git then re-run the make, but then we could just add into the script to re-pull the new command (as I have suggested in 126.

There is also a lot of discussion as to whether you should use make in this way but that is always the case with open source.

perrin4869: ah right, I see what you mean, I didn't realize the install shell could be executed that way due to tunnel vision 😅
I guess where I'm coming from is, I don't really want to install programs system-wide outside of using my OS package manager as much as possible, and the makefile would give extra flexibility in choosing how to install zwift

perrin4869: as for the nix flake, doesn't it just apply no nixos?

sHedC: > ah right, I see what you mean, I didn't realize the install shell could be executed that way due to tunnel vision 😅 I guess where I'm coming from is, I don't really want to install programs system-wide outside of using my OS package manager as much as possible, and the makefile would give extra flexibility in choosing how to install zwift

Could just add an option to install the script to copy it globally or locally to /usr/local/bin or to ~/bin (assming the second is in your path which in Fedora it is.)

Docker is always root (you need the docker group to run it without sudo), not sure if you can make it userspace don't use it.
Podman is by default userspace (unless you decide to change it to run as root)

sHedC: > as for the nix flake, doesn't it just apply no nixos?

Seems it is a package or nixos, but I have only glanced at the docs briefly.

netbrain: Nix is a package manager, nixos is a operating system using nix as its package manager. 

Nix provides reliable reproducibility of packages, even more so than i.e a Dockerfile 

The nix flake was ment to provide a means to create several types of output of the Zwift package. One of them could be a flatpak build as an example.

sHedC:
Shall I close this then and focus on nix build.

netbrain:
I'm busy this weekend, but I'll take another look at it next week and see how we best can fit this idea into the project. There is nothing wrong with mixing and matching different tools like make, nix, bash scripts. Just want to make sure we move in the best direction.

netbrain (changes requested):
I don't think makefile is the way to go for the current setup of the project. Feel free to try to convince me, but we really don't need multiple ways of installing the required files. One should suffice. 

And for now, ive opted in to doing it like how kubectl and minikube + plenty others do it:

- https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/
- https://minikube.sigs.k8s.io/docs/start/
- https://github.com/docker/docker-install/ (even docker!)

The reason behind this is that it's supposed to be dead simple for a user.  with as little effort and maintenance required from me or any other contributor.

So for a users perspective, there is little or no barrier to entry, simply copy paste a shell command and voila, zwift is per-se "installed". (given that they have curl/wget or whatever)

However with the makefile you need to checkout the entire project (which shouldn't be neccesary) and then issue make xxx in order to install.

So it's a little more convoluted, it might also require users to install make if the system they are using doesn't come pre-installed with it. Same goes for git, or zip, depending on how you download the repository contents.

However i do welcome the zwift desktop and logo file if you still wish to contribute this. These changes have to be incldued in the install.sh script in that case.

Alltough now when i think about it, it might just be better if we'd provide a flatpak for the gui heavy users and a install.sh for the terminal heavy users :)

perrin4869:
ok, I removed the makefile.
I still think it provides the most intuitive interface to get the files installed into the destination that you want, with the current `install.sh` it would take more work to get different `PREFIX` type variables, etc, but if the idea is to always install them as system files into `/usr/local`, this should be enough

netbrain:
How about adding a uninstall flag like this?

```
#!/usr/bin/env bash

# Check if running as root
if [ "$(id -u)" -ne 0 ]; then
    echo "Please run as root"
    exit 1
fi

install_zwift() {
    echo "This will install zwift.sh into /usr/local/bin"
    read -p "Press enter to continue"

    mkdir -p /usr/local/bin
    curl -s -o /usr/local/bin/zwift https://raw.githubusercontent.com/netbrain/zwift/master/zwift.sh
    chmod +x /usr/local/bin/zwift

    mkdir -p /usr/local/share/icons/hicolor/scalable/apps
    curl -s -o /usr/local/share/icons/hicolor/scalable/apps/Zwift\ Logogram.svg https://raw.githubusercontent.com/netbrain/zwift/master/assets/hicolor/scalable/apps/Zwift%20Logogram.svg

    mkdir -p /usr/local/share/applications
    curl -s -o /usr/local/share/applications/Zwift.desktop https://raw.githubusercontent.com/netbrain/zwift/master/assets/Zwift.desktop

    if [[ ":$PATH:" != *":/usr/local/bin:"* ]]; then
        echo "WARNING: Could not find /usr/local/bin on the \$PATH, you might need to add it."
    fi
}

uninstall_zwift() {
    echo "This will remove zwift.sh and associated files"
    read -p "Press enter to continue"

    rm -f /usr/local/bin/zwift
    rm -f /usr/local/share/icons/hicolor/scalable/apps/Zwift\ Logogram.svg
    rm -f /usr/local/share/applications/Zwift.desktop

    echo "Zwift has been uninstalled."
}

if [ "$1" == "uninstall" ]; then
    uninstall_zwift
else
    install_zwift
fi
```

And also add this to the README?

Also, in terms of packages like deb, rpm, aur or whatnot. Yes sure that would be interesting. But i still think this repo should be the contents of the "source" files just like for instance wine, and then have independent package maintaners that do the actual creation of deb's, rpm's in their own repositories. What do you think?

Feel free to make your own aur repository, there you can add your own customizations as you see fit.

sHedC:
> How about adding a uninstall flag like this?
> 
> ```
> #!/usr/bin/env bash
> 
> # Check if running as root
> if [ "$(id -u)" -ne 0 ]; then
>     echo "Please run as root"
>     exit 1
> fi
> 
> install_zwift() {
>     echo "This will install zwift.sh into /usr/local/bin"
>     read -p "Press enter to continue"
> 
>     mkdir -p /usr/local/bin
>     curl -s -o /usr/local/bin/zwift https://raw.githubusercontent.com/netbrain/zwift/master/zwift.sh
>     chmod +x /usr/local/bin/zwift
> 
>     mkdir -p /usr/local/share/icons/hicolor/scalable/apps
>     curl -s -o /usr/local/share/icons/hicolor/scalable/apps/Zwift\ Logogram.svg https://raw.githubusercontent.com/netbrain/zwift/master/assets/hicolor/scalable/apps/Zwift%20Logogram.svg
> 
>     mkdir -p /usr/local/share/applications
>     curl -s -o /usr/local/share/applications/Zwift.desktop https://raw.githubusercontent.com/netbrain/zwift/master/assets/Zwift.desktop
> 
>     if [[ ":$PATH:" != *":/usr/local/bin:"* ]]; then
>         echo "WARNING: Could not find /usr/local/bin on the \$PATH, you might need to add it."
>     fi
> }
> 
> uninstall_zwift() {
>     echo "This will remove zwift.sh and associated files"
>     read -p "Press enter to continue"
> 
>     rm -f /usr/local/bin/zwift
>     rm -f /usr/local/share/icons/hicolor/scalable/apps/Zwift\ Logogram.svg
>     rm -f /usr/local/share/applications/Zwift.desktop
> 
>     echo "Zwift has been uninstalled."
> }
> 
> if [ "$1" == "uninstall" ]; then
>     uninstall_zwift
> else
>     install_zwift
> fi
> ```
> 
> And also add this to the README?
> 
> Also, in terms of packages like deb, rpm, aur or whatnot. Yes sure that would be interesting. But i still think this repo should be the contents of the "source" files just like for instance wine, and then have independent package maintaners that do the actual creation of deb's, rpm's in their own repositories. What do you think?
> 
> Feel free to make your own aur repository, there you can add your own customizations as you see fit.

Yes I thought that package maintainers take a repo like this and alter to suite their needs, at least that is how I understood how it works.

perrin4869:
Done! Please note that if using `curl` to execute the script, passing the `uninstall` parameter is not straightforward.

Well, if there is no `Makefile` available, and someone wants to make a `SlackBuild` or `AUR` build script, then yeah, they will just figure out where to put the relevant files and package the program as they think is best. Note that this isn't always obvious, for example, where to place the icon and desktop files, and it could lead to inconsistent packaging in different repos, and people duplicating efforts that could be centralized in a makefile, to have a standard that everyone else can follow. I think if there is a desire to see a program land on AUR, etc, it is a good idea to have a Makefile for those reasons

sHedC:
> Done! Please note that if using `curl` to execute the script, passing the `uninstall` parameter is not straightforward.
> 
> Well, if there is no `Makefile` available, and someone wants to make a `SlackBuild` or `AUR` build script, then yeah, they will just figure out where to put the relevant files and package the program as they think is best. Note that this isn't always obvious, for example, where to place the icon and desktop files, and it could lead to inconsistent packaging in different repos, and people duplicating efforts that could be centralized in a makefile, to have a standard that everyone else can follow. I think if there is a desire to see a program land on AUR, etc, it is a good idea to have a Makefile for those reasons

Could we make uninstall/ an option e.g. if installed then ask to re-install or uninstall with re-install being the default.
or move to two scripts to make copy paste easier?

But if we do the select in the script then would want a bypass as in zwift I have a self updating part.

perrin4869:
Please feel free to improve on the `install` script, I'm not really interested much in it personally 😅

perrin4869:
hm... maybe it's best to take the `uninstall` part into a separate PR? Or feel free to modify this one, sorry, I'm not really interested in improving it 😅

netbrain:
Yeah, lets just keep it as is for now. We can in future include uninstall instructions in readme and/or provide a script for it.

perrin4869:
not to be super persistent, but I think that there were some misconceptions about how package managers usually work
package managers rarely will apply changes to the source directory, this is a last-resort measure. The reason is that there are no guarantees that the changes will remain compatible with future versions, and an intent to keep the package as close as possible to the intents of the author

please take a look at one build script from AUR: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=tofi

You can see that this script lists the source file as a `tar.gz` link to github. This package will then be downloaded, unpacked, and no modifications will be applied to it. You can see also that there is the `package` function, where the built files will be installed into a predetermined directory, and that directory will go on to be packaged for distribution.

Usually packages that are meant for distribution via deb, rpm, etc, will have some means to create their directory structure inside their source files, it won't usually be left as an exercise to the package maintainers. In the case above, it is using `mason` and `ninja`, in other cases it's `autotools` and sometimes it is just a `makefile`. Just thought I would try to clear that up.

sHedC:
> not to be super persistent, but I think that there were some misconceptions about how package managers usually work package managers rarely will apply changes to the source directory, this is a last-resort measure. The reason is that there are no guarantees that the changes will remain compatible with future versions, and an intent to keep the package as close as possible to the intents of the author
> 
> please take a look at one build script from AUR: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=tofi
> 
> You can see that this script lists the source file as a `tar.gz` link to github. This package will then be downloaded, unpacked, and no modifications will be applied to it. You can see also that there is the `package` function, where the built files will be installed into a predetermined directory, and that directory will go on to be packaged for distribution.
> 
> Usually packages that are meant for distribution via deb, rpm, etc, will have some means to create their directory structure inside their source files, it won't usually be left as an exercise to the package maintainers. In the case above, it is using `mason` and `ninja`, in other cases it's `autotools` and sometimes it is just a `makefile`. Just thought I would try to clear that up.

Its fine, I got that wrong then, I follow a lot of arguments where maintainers are changing packages and the original maintainers are refusing to deal with issues relating to say xxx distro.

I will check that out.

PR #132, opened 2024-05-16
ShedC/issue126

sHedC:
Updates:

1. Re-factored zwift to try and make it more readable in sections
2. Added a 5 second timeout and question to auto update zwift if out of date
3. Added hostname inline with latest push to main
4. Added check for xhost
5. Moved wayland add to registry entry to build otherwise does not work (only activates if DISPLAY is unset)
6. Moved DISPLAY unset to entry point as doesn't work in run_zwift.sh
7. Merged Wayland and X11 flags into WINDOW_MANAGER and WM_FLAGS
8. Changed getting latest version to not cache (so don't pick up old versions) and removed warning about null pointer

Thats it, has conflicts but only because I didn't merge the hostname change and added it manually to mine.

### DIFF
--- a/zwift.sh
+++ b/zwift.sh
@@ -1,38 +1,61 @@
 #!/usr/bin/env bash
-set -x
+if [ ! -z $DEBUG ]; then set -x; fi
 
-### FORMATTING ###
-RED='\033[0;31m'
-NC='\033[0m'
-BOLD='\033[1m'
-UNDERLINE='\033[4m'
+DONT_PULL=1
+IMAGE=localhost/zwift
 
-### DEFAULT CONFIGURATION ###
+# MSG BOX use Zenity if available.
+msgbox() {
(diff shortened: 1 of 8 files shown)


zwift.sh
```diff
@@ -1,41 +1,57 @@
...
-VERSION=${VERSION:-latest}
-
-# Use podman if available
-if [[ ! $CONTAINER_TOOL ]]
-then
-    if [[ -x "$(command -v podman)" ]]
-    then
+if [ ! -z $DEBUG ]; then set -x; fi
+
+# MSG BOX use Zenity if available.
```
netbrain: remove comment?

netbrain (approved):

zwift.sh
sHedC: Good spot

sHedC: Completed.

PR #138, opened 2024-05-22
wrong desktop environment detected when starting x11 with startx

### LINKED ISSUE #137: wrong desktop environment detected when starting x11 with startx
This issue was introduced in https://github.com/netbrain/zwift/pull/132
It is now using `XDG_SESSION_TYPE` to detect the running window manager type, but this is not a reliable method. For example, in my current x11 session, `XDG_SESSION_TYPE` is set to `tty`. I imagine this is because I boot into tty, and use `startx` to start `x11`. In this case, the necessary `XAUTHORITY` environment variables and volumes will not be properly passed to the container and zwift will not launch. Probably reverting to the previous method will give a more reliable result.

sHedC:
Fixes #137

### DIFF
--- a/zwift.sh
+++ b/zwift.sh
@@ -71,9 +71,26 @@ fi
 
 ########################################
 ###### OS and WM Manager Settings ######
-if [  "$XDG_SESSION_TYPE" == "wayland" ]; then
+case "$XDG_SESSION_TYPE" in
+    "wayland")
+        WINDOW_MANAGER="Wayland"
+    ;;
+    "x11")
+        WINDOW_MANAGER="XOrg"
+    ;;
+    *)
+        if [ ! -z $WAYLAND_DISPLAY ]; then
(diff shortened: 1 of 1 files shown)


sHedC:
One sec just noticed something else and infact now XWayland and XOrg are very close so merging.

sHedC:
Comment here, on the WM check I now that XDG is probably now overkill but as it fixes some issues with other distros and window managers that the second check shoudl also handle.

Shall I risk removing it and sticking to jsut this?  Just don't want to risk any breaking for users it currentlly works for.

```
if [ ! -z $WAYLAND_DISPLAY ]; then
    WINDOW_MANAGER="Wayland"
elif [ ! -z $DISPLAY ]; then
    WINDOW_MANAGER="XOrg"
fi
```

netbrain (approved):
lgtm

PR #141, opened 2024-05-29
Update to Wine 9.9 development to support wayland native.

sHedC:
Update to Wine 9.9 development to support wayland native.

Move Config check ealier to allow for env settings in config file better.

Fix Icon showing intask bar when started from menu

Allow wine version fix in docker build (fixed to 9.9)


update_zwift.sh
```diff
@@ -58,11 +58,17 @@ function wait_for_zwift_game_update() {
 
 if [ ! "$(ls -A .)" ] # is directory empty?
 then
+    # Prevent Wine from trying to install a different mono version
+    WINEDLLOVERRIDES="mscoree,mshtml=" wineboot -u
```
netbrain, line 62: :partying_face:

zwift.sh
```diff
@@ -36,6 +36,24 @@ msgbox() {
...
+    source $HOME/.config/zwift/config
+fi
+
+# Check for $USER specific zwift configuration, sourced here and passed on to container aswell
+if [[ -f "$HOME/.config/zwift/$USER-config" ]]
+then
+    ZWIFT_USER_CONFIG_FLAG="--env-file $HOME/.config/zwift/config"
+    source $HOME/.config/zwift/config
+fi
+
```
netbrain: Guess this needs to be updated/rebased from #142

netbrain (approved):
Lgtm

update_zwift.sh
sHedC: Not sure if we even need mono to be installed at all?  As we install dot.net which overrides it right?

zwift.sh
sHedC: Done

update_zwift.sh
netbrain: no wine mono is required atleast it was in all my previous testing.

netbrain (approved):

netbrain:
Ill merge this and build zwift from scratch which will from this point onward create a new zwift image with wine 9.9.

:crossed_fingers:

netbrain:
https://github.com/netbrain/zwift/actions/runs/9299444600

update_zwift.sh
sHedC: I will try

sHedC:
I'll try without mono I think it was needed before because it auto installs on wineboot.  But will  check :)

sHedC:
Build failed I ahve seen that before it is getitng a previous version from the web site and updating to the current version in this case zwift is 1.65 from the website and 1.66 is the actual.  

I updated the line to not use cache but still :(

sHedC:
I am updateing to add a better check that checks that one version is equal to or greater than

PR #146, opened 2024-06-06
Zwift does not run if your UID does not match the container UID.

sHedC:
Update README about podman support.

Fix Zwift to use the right id's for podman if the Host UID does not match the default container ID.


netbrain:
@hobeone  :point_up:

hobeone:
Tracking through this I think it makes sense and does look like a no-op for docker.  I'm travelling for the next few weeks so can't fully test myself but in limited testing it still seems to work for docker.

netbrain (approved):

PR #148, opened 2024-07-04
fix(screensaver): detect xscreensaver

perrin4869:
the ever evolving battle against screensavers (or in favor of screensavers, depending how you look at it) rages on.
turns out that I was missing `xscreensaver-command` which is [used by xdg-screensaver](https://gitlab.freedesktop.org/xdg/xdg-utils/-/blob/v1.1.3/scripts/xdg-screensaver.in?ref_type=tags#L885) to detect and inhibit xscreensaver.
this explains why I was still getting screensavers while zwift was open, whenever youtube would not be playing on my other screen. oops 😅
I think  a few more environment variables might be needed to make available to detect other screensavers or power management schemes in gnome/kde, but I don't use those systems so I couldn't test them. I could try to add those based on the contents of `xdg-screensaver` though

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,9 +40,10 @@ RUN dpkg --add-architecture i386
 # - libvulkan1 for vulkan loader library
 # - procps for pgrep
 # - xdg-utils for xdg-screensaver
+# - xscreensaver for xscreensaver-command, used by xdg-screensaver to inhibit xscreensaver
 
 RUN apt-get update
-RUN apt-get install -y wget curl sudo winbind libgl1 libvulkan1 procps gosu xdg-utils
+RUN apt-get install -y wget curl sudo winbind libgl1 libvulkan1 procps gosu xdg-utils xscreensaver
 RUN wget -qO /etc/apt/trusted.gpg.d/winehq.asc https://dl.winehq.org/wine-builds/winehq.key
 RUN DEBIAN_VERSION=${DEBIAN_VERSION} echo "deb https://dl.winehq.org/wine-builds/debian/ ${DEBIAN_VERSION} main" > /etc/apt/sources.list.d/winehq.list
 RUN apt-get update


perrin4869:
please note that `xscreensaver-command -version` will fail with an error if no `xscreensaver` daemon is running on the current xorg session, so just making it available in the container won't give a false positive

sHedC:
That could get complex, we could rely on gamemoderun in the startup script?

https://github.com/FeralInteractive/gamemode

Maybe that would work as it is supposed to deal with that not sure about power I don't use power management on my desktop and laptop when plugged in its set to 4 hours (never cycled that long too boring).

Fedora install is just dnf install gamemode

????

perrin4869:
Hm... It's the first time I hear of this project!
From skimming though, it doesn't sound like it handles screensaver inhibition?

sHedC:
> From skimming though, it doesn't sound like it handles screensaver inhibition?

It does, its part of what it is supposed to do its in the readme that it inhibits screen saver.

image

on a side note I think this is a fine submission from my point of view, nothing problematic I see with it.  I will investigate gamemode as it should be possible to build it into the docker as there are games and flatpaks with it built in.

sHedC (approved):
No issues with this

perrin4869:
Oh right sorry!
Interesting, would need to evaluate exactly what it's doing to inhibit the screen. Since every screensaver has its own idiosyncrasies as to how to inhibit the screen, I'd be surprised if it covers as many cases as xdg-screensaver

sHedC:
@perrin4869 - I tried gamemode inside the docker worked for me not sure if you want to try:

In Dockerfile add game mode to line 45
```RUN apt-get install -y wget curl sudo winbind libgl1 libvulkan1 procps gosu xdg-utils gamemode```

In run_zwift.sh - comment out the xdg screen saver and add gamemoderun
```
[[ -n "${DBUS_SESSION_BUS_ADDRESS}" ]] && while true; do sleep 30; xdg-screensaver reset; done &

echo "Killing uneccesary applications"
pkill ZwiftLauncher
pkill ZwiftWindowsCra
pkill -f MicrosoftEdgeUpdate

/usr/games/gamemoderun wineserver -w
```

For me this enabled gamemode (I have a gnome extension to show when game mode is running.  It stopped the screen saver which I set to 2 min.

perrin4869:
thanks!
ok, just took a look at `gamemode` source code, the inhibiting is [done here](https://github.com/FeralInteractive/gamemode/blob/a2fe0108b59948ed861ccc087aa91af9273b036f/daemon/gamemode-dbus.c#L733)
I haven't tried it yet, but I think this would work in my personal use-case (running the `xscreensaver-systemd` daemon). Also, note that this is also the method used by `firefox` when running youtube videos, for example.
@sHedC what screensaver were you using? I think this would work also on recent versions of gnome and kde, but browsing through `xdg-screensaver`, it seems systems like `mate` and `cinnamon` and old versions of `gnome` and `kde` are not using the `freedesktop` dbus interface, [for example](https://gitlab.freedesktop.org/xdg/xdg-utils/-/blob/master/scripts/xdg-screensaver.in?ref_type=heads#L886). But it would be worth trying on those systems as well, `cinnamon` and `mate` did add `freedesktop` support in the meantime to their screensavers.
And then there are the other implications of running `gamemode` to consider?
I'll try this later (or maybe tomorrow) for sure to confirm that it works, as of right now I think it would be fine replacing `xdg-utils` and with `gamemode` if support for those systems is question is not a priority

sHedC:
I am using Gnome Fedora 40 and KDE fedora 40 usually test with both wayland and X as its an environment just for building and testing zwift.  Also have a NixOS build with hyperland but don't really get that OS :)

I actualy only know about it because of steam never thought it inhibited screen saver but was using it for gaming.

sHedC:
Gamemode changes to performance mode and changes some CPU scheduling if it can.  It can't always do that it checks what it can do before it does.

But it is used by default in some games now such as tomb raider.

sHedC:
Up to you if you want to update on this PR please do I will test it on some others, but basically game mode is used by Steam so I expect good support.

If not I can create a branch and do it there to test under a new issue/ PR :)

sHedC:
Just as a note quick google and game mode seems to be working on a lot of distros and windows managers

sHedC:
Closing in favour of #149 to use gamemode.

PR #149, opened 2024-07-05
feat: use gamemode for screensaver inhibit

perrin4869:
This is an alternative to https://github.com/netbrain/zwift/pull/148
The main concern is that this may not work for someone running mate or cinnamon, but can't tell for sure without trying it out or digging into their source code...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,9 +40,10 @@ RUN dpkg --add-architecture i386
 # - libvulkan1 for vulkan loader library
 # - procps for pgrep
 # - gamemode for freedesktop screensaver inhibit
+# - xdg-utils seems to be a dependency of wayland
 
 RUN apt-get update
-RUN apt-get install -y wget curl sudo winbind libgl1 libvulkan1 procps gosu gamemode
+RUN apt-get install -y wget curl sudo winbind libgl1 libvulkan1 procps gosu gamemode xdg-utils
 RUN wget -qO /etc/apt/trusted.gpg.d/winehq.asc https://dl.winehq.org/wine-builds/winehq.key
 RUN DEBIAN_VERSION=${DEBIAN_VERSION} echo "deb https://dl.winehq.org/wine-builds/debian/ ${DEBIAN_VERSION} main" > /etc/apt/sources.list.d/winehq.list
 RUN apt-get update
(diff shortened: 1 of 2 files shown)


sHedC:
I will  run tests to day on a few of my VM's. Learning how to checkout and test PR's :)

sHedC:
have an issue with Wayland but not sure what it is yet so testing.

sHedC (changes requested):
In the Docker File, Wine running experimental wayland needs the xdg-utils installed, can you re-add that component for Wine Wayland please.

perrin4869:
Wow that's interesting! Thanks for testing 😀 
I'll update the PR tomorrow
Btw, do you know which program from xdg-utils is needed by experimental Wayland? Wonder why it isn't a dependency of the package

sHedC:
Um no i will see if I can find out. It just complains that the wrong
version of opengl so will check the wine forum

sHedC:
as far as I can tell it uses xdg-open but maybe also xdg-runtime.

sHedC:
Testing

- [ ] Fedora 40 Gnome X11
- [ ] Fedora 40 Gnome Wayland (both X11 and Wayland Experimental)
- [X] Mint 21 - Cinnamon (whatever is default)
- [ ] Nobara 39 KDE (Wayland)

sHedC:
Tested what I can seems to work fine, let me know when you have re-added xdg-utils.

perrin4869:
Ok, added `xdg-utils` back. I'm wondering if maybe this is a bug in the dependencies on the Ubuntu package side though

sHedC:
> Ok, added `xdg-utils` back. I'm wondering if maybe this is a bug in the dependencies on the Ubuntu package side though

I think that wine wayland is depending on xdg-shell maybe, but then I am not that into wine dev to know, I did find this which is an read: https://bugaevc.gitbooks.io/writing-wayland-clients/content/beyond-the-black-square/xdg-shell.html

sHedC (approved):
Looks good for me

sHedC:
@netbrain - I have reviewed and tested this seems to be good for both screen saver/ power management and performance in all my tests.

netbrain (approved):

netbrain:
Looks good! Great work

Issue #147, opened 2024-07-03
Fedora 40 with Nvidia - Performance


cmuench:
**Describe the bug**
If I start zwift on my Fedora Linux machine the GPU is not used.
Maybe a regression bug.
I know that the NVIDIA_DRIVER_CAPABILITIES is set in the Dockerfile.

**Expected behavior**
I can use my Nvidia GPU

**Additional context**
If  I add `-e NVIDIA_DRIVER_CAPABILITIES=all `  the problem is fixed.

That's my patched file:

```
#########################
##### RUN CONTAINER #####
CONTAINER=$($CONTAINER_TOOL run ${GENERAL_FLAGS[@]} \
        $ZWIFT_CONFIG_FLAG \
        $ZWIFT_USER_CONFIG_FLAG \
        $VGA_DEVICE_FLAG \
        -e NVIDIA_DRIVER_CAPABILITIES=all \
        ${DBUS_CONFIG_FLAGS[@]} \
        ${WM_FLAGS[@]} \
        ${PODMAN_FLAGS[@]} \
        $IMAGE:$VERSION $@
)
```

sHedC:
Are you running the nvidia driver or Nouveau?  I assume Nvidia?

cmuench:
Proprietary nvidia driver. I am on the 555 which is really fresh.

sHedC:
I'll look into it currently using Nouveau as not got round to setting up the proprietary but have to for another project.

cmuench:
No worry. I am able to fix this issues from my side. I opened this issue for other people with the same problem.
I just had a view to the Dockerfile.

The ENV var was introduced there.
https://github.com/netbrain/zwift/blob/cf43c78c8fb4e154002f9107be6f30705e9d295b/Dockerfile#L29

In the meantime we have multi step build there.
In the last step the ENV var is missing.
If we move the env var after the `FROM wine-base` it may be fixed.

sHedC:
If you can give that a go :) if not I will try it out once I get my driver switched over.

sHedC:
Hi, I switched to the nvidia drivers, after some issues with resiable bar and the /dev/card changing sorted.  I don't have this issue but I only have an NVIDIA card in mine no other capable device.

Do you have an intel chip with video?  
And do you think it might be related to #21

cmuench:
@sHedC I have a AMD based system (AMD Ryzen 9 7950X) with a  NVIDIA 4060 TI.

[...]

Issue #89, opened 2024-04-11
Fix the "How can I add custom .zwo files?" section in the README


netbrain:
Improve this section in the README and check if it is currently correct.

Simplify if possible.

netbrain:
@jonathanglima :point_up:  not sure if the latest wine 9 has broken this for you. But i think the paths may be incorrect now. And maybe this could be simplified to allow for a separate volume for zwo files. Depends on where the zwo files are located.

sHedC:
Would doing a script to add or make an option for zwift to make use of that to copy or add a zwo mount point?  then get the id from the auth and do a link or sync from the mount point to the Workouts?

e.g.  `ZWIFT_ZWO=/home/<user>/.config/Zwift/Workouts zwift`
or just use the config area to say if exists Workouts then mount to copy/ link?

Looking up the ID from the login details is as follows.
`curl -sSl --get --data-urlencode "username=user" --data-urlencode "pw=password" https://z00pbp8lig.execute-api.us-west-1.amazonaws.com/latest/zwiftId`

Thoughts?

netbrain:
Probably better if we provide a way to find your Zwift Id as a helper script or just in the docs and then have two env vars.

ZWIFT_ZWO=/home/netbrain/my-zwo
ZWIFT_ACCOUNT_ID=1234

With these two we can link or copy the files into the container. Preferably link, but then you must link with proper uid/gid.

netbrain:
I try to follow KISS (keep it simple stupid) whenever possible

sHedC:
Added Support in latest, see the README.

PR #150, opened 2024-07-08
Update to Support zwo files better and other Minor Fixes.

sHedC:
Some additional fixes below, this requires a rebuild of the docker.

- #89 Added the abiltiy to map a directory for custom zwo files to a local directory using an environment parameter, updated readme.
- Minor Fix to Docker to make the entrypoint and other files rx by default (from just x)
- Minor Fix for Podman Build to set ipc host temporarily, can cause failures on laptops
- #147 - Attempt to fix issue by moving the Enviornment Variables for NVIDIA
- Update Readme to point to new Nvidia-container-toolkit not the deprecated one

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,9 +25,6 @@ ARG WINE_VERSION="=9.9~trixie-1"
 
 ARG WINETRICKS_VERSION=20240105
 ARG DEBIAN_VERSION
-ENV NVIDIA_VISIBLE_DEVICES=all
-ENV NVIDIA_DRIVER_CAPABILITIES=all
-ENV WINEDEBUG=fixme-all
 
 RUN dpkg --add-architecture i386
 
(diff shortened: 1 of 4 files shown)


netbrain (approved):

netbrain:
Looks good. 👍

sHedC:
Added minor update to the ZWO section to talk about the subfolder and internal zwift id.

sHedC:
@netbrain -  FYI, Build is failing since last night due to some dependency in Wine that is blocked.  Stable and Devel branches.

netbrain (approved):

netbrain:
Ok, you need anything from me?

sHedC:
> Ok, you need anything from me?

No its broken at the wine side, there is already an issue related.  Just informing as you can't merge this request because build fails.  I think something to do with related updates where the 64 and 32 bit library versons are conflicting.

I noticed as I was testng a new zwift builder workflow started failing last night.

Once fixed I will merge and build :)

sHedC:
@netbrain - Trixie build still borked, due to issues between the 386 and 64 dependancies.  It builds and runs with Bookworm but Nvidia using Experimental Wayland is stuffed in bookworm (no idea why).

Guess either wait for trixie to be fixed or note that Nvidia can't run in Wayland experimental mode.  To be fair Nvidia under wayland is not good anyway as eventually the NVRAM fills and programs crash.

PR #164, opened 2024-10-02
zwift-auth: fix invalid variable in latest curl

quietvoid:
Starting from `curl` version `8.9.0`, `REDIRECT_URL` in caps is not recognized.
Currently the latest `netbrain/zwift` image is still using `curl` version `8.8.0` so it's still working but if it were built from scratch, it would get an updated `curl`.


netbrain (approved):

netbrain:
Superb! Thanks alot 👍

PR #174, opened 2024-11-07
Make install script work with restrictive umask

stpierre:
If a user has set `umask 077` or other umasks more restrictive than `022` (the default on many distros), the install script will not work because it only enables the execute bit on `/usr/local/bin/zwift` for users for whom the file is already readable. This makes it set all permissions, rather than modifying the existing permissions, to 0755 (read/write/execute for `root`, read/execute for others).


netbrain (approved):
Thanks! looks good.

PR #178, opened 2024-11-17
Support storing password in SecretService keyring

Jmennius:
That should be a bit safer that storing the password in plain text, at least for my OCD 😄 
Needs testing with docker (I've tested with podman).

- [ ] test with docker

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,14 @@ The credentials will be used to authenticate before launching the Zwift app, and
 
 Note: This will be loaded by zwift.sh in cleartext as environment variables into the container.
 
+Alternatively, instead of saving your password in the file, you can save your password in the SecretService keyring like so:
+
+```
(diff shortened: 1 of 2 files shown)


netbrain (commented):
Looks good to me, @sHedC any input? 

My only concern is that the complexity we keep adding will make it undoubtedly harder to support and change in the future.

sHedC:
> Looks good to me, @sHedC any input?
> 
> My only concern is that the complexity we keep adding will make it undoubtedly harder to support and change in the future.

Yes, but I think it is probably a good idea for those people who do not encrypt drives.  But what about regression for those with password already set.

I will check this out and test against podman and docker.

Jmennius:
So I've recently learned that `docker secret` is not available in regular docker:
```
docker secret ls
Error response from daemon: This node is not a swarm manager. Use "docker swarm init" or "docker swarm join" to connect this node to swarm and try again.                                                               
```
This basically means that implementation for docker is broken at the moment.

netbrain:
Should probably add this clarification in the readme.

PR #199, opened 2025-06-08
Addition of CPU throttling configuration

rommeswi:
Adds an additional configuration option CONTAINER_CPUS.
Adds CPUS_FLAG to docker/podman command.
Updates readme

I didn't add anything to the NixOS configuration files. I worry I might break something.


netbrain:
It's not that difficult to add it to the flake.nix

Simply add the env as it has been done for the other config options. 

I'll review it after you have done so. Good job 👍!

netbrain:
@rommeswi I generalized your solution, would you agree with these changes? and could you test if this works for your usecase? Makes it a bit more flexible, and can be used for more than just --cpus.

rommeswi (commented):
These changes look great! Indeed, it makes a lot of sense to allow for use of other flags as well and handle all of these in one argument.

rommeswi:
I just tested the new version and it runs flawlessly! Thank you!

PR #201, opened 2025-06-22
Correct paths for `$USER-config`

netbrain:
Came across this fork with this fix, should implement this.

Thanks @bluesquall


netbrain:
Nevermind, seems like it's already implemented :confused:

PR #212, opened 2025-09-19
feat: add ZWIFT_PROFILE_DIR config option

kschat:
Zwift was relegating me to medium settings despite having a Ryzen 7950X and RX 6800. `ZWIFT_PROFILE_DIR` allows users to mount the directory which contains the graphics profiles to the host machine without removing the existing files from the directory in the container.


netbrain (approved):
Looks good to me! Thanks 🙏👍

PR #213, opened 2025-09-19
fix: update ZWIFT_FG to just disable setting "-it"

kschat:
I was trying to setup Steam as the launcher for Zwift and found that, although it launched Zwift, Steam acted as though the process exited right away. By updating `ZWIFT_FG` to not set `-d` rather than replace `-d` with `-it`, Steam was able to detect that the process was still running. 

Since `INTERACTIVE` sets `-it` I figured this was a safe change since users can use that to get an interactive TTY, but if you think otherwise let me know. I'm happy to add a new config option and revert `ZWIFT_FG` to the previous behavior instead.


README.md
```diff
@@ -205,7 +205,7 @@ By default, Zwift loads a set of graphics settings based on the selected game re
 
 You can find more information about these settings in this [Zwift Insider](https://zwiftinsider.com/config-file-tweaks/) article.
```
kschat, line 207: Fix a typo from my previous PR :man_facepalming:

netbrain (approved):

Issue #214, opened 2025-09-20
ZWIFT_PROFILE_DIR does not work with podman


glennvl:
The arguments used to mount the graphics profiles directory do not work with podman.

```sh
ZWIFT_PROFILE_DEST="/home/user/.wine/drive_c/Program\ Files\ \(x86\)/Zwift/data/configs"
ZWIFT_PROFILE_VOL="--mount dst=$ZWIFT_PROFILE_DEST,volume-opt=device=$ZWIFT_PROFILE_DIR,type=volume,volume-driver=local,volume-opt=type=none,volume-opt=o=bind"
```

Why not use the same way of mounting as the workout, activity and log directories?

```sh
ZWIFT_PROFILE_VOL="-v $ZWIFT_PROFILE_DIR:/home/user/.wine/drive_c/Program\ Files\ \(x86\)/Zwift/data/configs"
``` 

Or alternatively

```sh
ZWIFT_PROFILE_VOL="--mount type=bind,src=$ZWIFT_PROFILE_DIR,dst=/home/user/.wine/drive_c/Program\ Files\ \(x86\)/Zwift/data/configs"
```

In the documentation there is also a note about zwift automatically creating the profiles if the directory is empty. That did not happen for me. I manually copied the profiles from the container and replaced the contents of the medium.txt file with the contents of the ultra.txt file.

```text
NOTE: The profiles are generated by Zwift, so if you're directory is empty, start Zwift once to generate the profiles.
```

glennvl:
I also have a somewhat related question about the config file.

I can do for example `CONTAINER_EXTRA_ARGS="-v ~/.config/zwift/profiles/ultra.txt:/home/user/.wine/drive_c/Program\ Files\ \(x86\)/Zwift/data/configs/medium.txt" zwift`. But if I add the `CONTAINER_EXTRA_ARGS` to the config file, it errors out (weird stuff happens escaping the backslashes etc). What is the correct way to add an option with spaces in the config file?

kschat:
> The arguments used to mount the graphics profiles directory do not work with podman.

Sorry, I don't use podman. What error do you get? 

> Why not use the same way of mounting as the workout, activity and log directories?

Because that would overwrite the directory instead of copying the existing content into the host directory.

> In the documentation there is also a note about zwift automatically creating the profiles if the directory is empty. That did not happen for me.

Was this with podman or docker?

> Also your vs you're.

Thanks! I fixed that in #213.

glennvl:
[...]

PR #215, opened 2025-09-21
Issue 214 - ZWIFT_PROFILE_DIR

glennvl:
Alternative implementation of `ZWIFT_PROFILE_DIR`. See #214.

### DIFF
--- a/run_zwift.sh
+++ b/run_zwift.sh
@@ -9,6 +9,11 @@ if [ ! -d "$ZWIFT_HOME" ]; then
   exit 1
 fi
 
+# Restore zwift graphics profiles if the directory is mounted and empty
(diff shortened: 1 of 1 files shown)


netbrain:
@kschat does this look good for you?

kschat:
Reviewing.

run_zwift.sh
```diff
@@ -9,6 +9,11 @@ if [ ! -d "$ZWIFT_HOME" ]; then
   exit 1
 fi
```
kschat, line 11: If the directory isn't empty, but only contains one of the profile files, then this will cause the other profile files to not be created in the host directory. I haven't tested this extensively in Zwift, but I would have to assume this would be an unexpected state for Zwift if you were to change profiles to one that doesn't exist and could cause a crash.

netbrain:
Yeah, i agree with @kschat that using the built in volume mounting functionality of docker and podman is probably the way to go for now.

PR #216, opened 2025-09-22
fix: ZWIFT_PROFILE_DIR for podman

kschat:
`volume-driver=local` is useful for when you're running Docker in a container (e.g. Docker Desktop), but that's only relevant when using a non-Linux host, which obviously isn't the case for this project. Removing this option works for Docker and Podman, I tested locally (sorry for not testing with both initially).


netbrain:
No worries, thanks for the contrib.!

Issue #205, opened 2025-07-29
Cannot find screenshots


rommeswi:
**Describe the bug**
I tried to find the screenshots taken in game. They did not appear in my pitures folder, document folder, download folder, workout folder, nor the config folder.

**To Reproduce**
1. Install on arch
2. take screenshots
3. try to find them

**Expected behavior**
The screenshots should appear in whatever the system's configured "Pictures" or "Screenshots" folder are. Alternatively, there should be a configuration option specifying the path for the screenshots.

**Screenshots**

I tried to take some, but could not find them!

netbrain:
You can now use the INTERACTIVE flag to start the container interactively, it should provide you with the means of finding these screenshots.

glennvl:
@netbrain Is this something we want to support out of the box? It could be as simple as mounting the configured pictures directory to the pictures directory in wine. Alternatively we could add another setting to map a screenshot directory, but I think in this case it probably makes sense to map to the pictures directory by default.

```bash
ZWIFT_PICTURES_VOL="-v $(xdg-user-dir PICTURES):/home/user/.wine/drive_c/users/user/Pictures"
```

If we want to do this, I can have a look at implementing it.

netbrain:
Sure thing, go for it!

glennvl:
@rommeswi The pull request was merged. You can now map a pictures directory, for example `ZWIFT_SCREENSHOTS_DIR=$(xdg-user-dir PICTURES)/Zwift` if you want the screenshots to be saved in your user pictures directory (Note that the directory has to exist, so you probably need to create the `Zwift` directory there first).

PR #222, opened 2025-09-30
Issue 205 access screenshots

glennvl:
This branch implements the features requested in #205.

* Map the user's pictures directory to the wine pictures directory where zwift stores screenshots
* Smuggled in a small fix for yesterday's pr :grimacing:

New screenshots taken in zwift will be saved in the `/home/$USER/Pictures/Zwift` directory.


zwift.sh
```diff
@@ -206,6 +206,9 @@ fi
 #############################
 ##### PREPARE ALL FLAGS #####
 
+USER_PICTURES_DIR="$(xdg-user-dir PICTURES)/Zwift"
```
netbrain: This opens up for a whole bunch of permission errors, especially if current user uid/gid != container user uid/gid. Would recommend this be a  named volume instead. like zwift-$USER is mapped, maybe zwift-pictures-$USER?

netbrain: alternatively, only map this as a folder mount if a config option is specified, so people who don't care about screenshots, can get the same experience.

glennvl: When creating a volume, the screenshots are still stored in a place that is not easily accessible, so that would not solve the issue of getting easy access to screenshots? To be able to access the zwift workouts directory we use a second mount as workaround. And is the permission issue not true for any file/directory that is mounted to the container?

If I understand it correctly, there is a permission problem when mounting files/directories if the user on the host has an id that is different from 1000 (the id of the user inside the container), which would be the case if for example there are multiple users on the host (first one gets id 1000, next user gets a different id)? In docker there is a workaround where the id of the container user/group are changed to match the id of the host user? And for podman the `--userns keep-id:uid=$CONTAINER_UID,gid=$CONTAINER_GID` is used? I'm sorry if I'm missing something obvious, but I'm trying to understand what still could go wrong.

netbrain: It might no longer be an issue, but i just remember back in the days (probably before --userns etc) we made the switch over to a named volume to circumvent issues regarding different uid/gid and file perms.

So we could try to merge, but there might be follow up changes required. But in any case, maybe not mount the pictures dir by default, let a config option enable this.

glennvl: @netbrain Reverted and added a config option instead.

PR #223, opened 2025-09-30
Use more secure container flags

### LINKED ISSUE #221: Why use --privileged and --security-opt label=disable?
Not exactly a bug but more of a security concern. 

Any reason why `--privileged` and `-security-opt label=disable` are used instead of more restrictive options? I could run the container with `--security-opt label=type:container_runtime_t` and `--device=/dev/dri` instead, which are better than running the container in privileged mode. 

The `entrypoint.sh` needs to check which containerization software is used, but this could be worked around by passing `CONTAINER` as an environmental variable from `zwift.sh` instead.

imjuzcy:
Removed the usage of `--privileged` and `--security-opt label=disable` by default to improve security of the container. `--security-opt label=type:container_runtime_t` is used instead. 

Fallback to insecure container with `--privileged` and `--security-opt label=disable` if the environment variable `PRIVILEGED_CONTAINER` is set to `1` for `zwift.sh`.

`--ipc host` flag is also removed from Podman build command, as it is also insecure.

Tested on Fedora 42 (amd64, SELinux Enforcing) with Podman 5.6.1. Builds and runs fine with the more secure options. 

More testing (with Docker and AppArmor systems) is likely needed before merge.

Fixes #221

### DIFF
--- a/.github/workflows/zwift_build_from_scratch.yaml
+++ b/.github/workflows/zwift_build_from_scratch.yaml
@@ -36,8 +36,8 @@ jobs:
           docker build -t netbrain/zwift:build .
           xhost +
           docker run --name zwift \
-                  --privileged \
                   -e DISPLAY=$DISPLAY \
+                  -e CONTAINER=docker \
                   --device /dev/dri \
                   -v /tmp:/tmp \
                   netbrain/zwift:build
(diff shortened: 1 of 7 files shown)


netbrain (approved):
LGTM

netbrain:
I think we can just merge this, and optionally revert or set PRIVILEGED_CONTAINER to 1 as default if there starts to pop up a lot of new issue submissions.

PR #224, opened 2025-10-01
Fix zwift.sh warning on line 72

glennvl:
The minor fixes from #222. But given that it will probably be a bit before that one gets merged, it's probably best to already merge these minor fixes first.

```console
./zwift.sh: line 72: [: : integer expression expected
```


netbrain:
agreed

PR #225, opened 2025-10-02
fix autologin by passing username into container

imjuzcy:
It seems that there was a mistake with autologin. The username was never passed into the container, although it checks for it. This should fix that.


netbrain (approved):
👍 looks legit! Must have been introduced during a refactor

PR #227, opened 2025-10-11
updated string to modern str variant + using inherit for repeated namespacing

argosnothing:
This addresses some of the build warnings on nix having to do with deprecated string module type usage. 

Also curious what format this flake is in. Was having issues formatting it properly.


netbrain:
This was contributed to the project by someone else. They will have to answer for that. 

In any case. Looks good for me.

PR #232, opened 2025-10-16
Lint bash scripts

glennvl:
Add a github action to lint all the bash scripts.

* Use <https://www.shellcheck.net/> as linter.
* Fix warnings in bash scripts.
* First run: <https://github.com/glennvl/zwift-linux/actions/runs/18561628658/job/52911822359>
* After fixing all warnings: <https://github.com/glennvl/zwift-linux/actions/runs/18566094688/job/52927488008>

I'm not sure if this is ok or if I'm out of line here since I had to make quite a bit of changes to the scripts to make the linter happy. The changes mostly have to do with the usage of double quotes and how conditions are written, there are no structural changes or changes to how everything works.

If we go through with this, which I personally think would be a good thing, I would like a thorough review of the changes first, preferably by multiple contributors. I also want to do further testing myself to make sure I did not unintentionally break anything before I consider this pull request ready to be merged.

---

### Status

TODO list until no longer draft

* [x] Sanity check with `podman`
  * [x] Build image
  * [x] Launch zwift
* [x] Test `build-image.sh` script still works
* [ ] Test `entrypoint.sh` script still works
  * [x] Test with `podman`
  * [ ] Test with `docker`
* [x] Test `install.sh` script still works
* [x] Test `run_zwift.sh` script still works
* [x] Test `update_zwift.sh` script still works
* [x] Test `zwift-auth.sh` script still works
* [ ] Test `zwift.sh` script still works
  * [x] All parameters work with `podman`
  * [ ] All parameters work with `docker`

### DIFF
--- a/zwift.sh
+++ b/zwift.sh
@@ -403,5 +403,5 @@ fi
 
 # Allow container to connect to X, has to be set for different UID
 if [ -n "$CONTAINER" ] && [ -x "$(command -v xhost)" ] && [ -z "$WINE_EXPERIMENTAL_WAYLAND" ]; then
-    xhost +local:"$($CONTAINER_TOOL inspect --format='{{ .Config.Hostname  }}' "$CONTAINER")"
+    xhost +local:"$($CONTAINER_TOOL inspect --format='{{ .Config.Hostname }}' "$CONTAINER")"
 fi


bin/build-image.sh
```diff
@@ -32,53 +29,42 @@ GENERAL_FLAGS=(
...
-$CONTAINER_TOOL run ${GENERAL_FLAGS[@]} \
-    $VGA_DEVICE_FLAG \
-    ${PODMAN_FLAGS[@]} \
-    $IMAGE:latest \
-    $@
-
-
-$CONTAINER_TOOL commit zwift $BUILD_NAME:latest
+$CONTAINER_TOOL build --force-rm -t "$BUILD_NAME" "$SCRIPT_DIR/../."
+$CONTAINER_TOOL run "${GENERAL_FLAGS[@]}" "${VGA_DEVICE_FLAG[@]}" "${PODMAN_FLAGS[@]}" "$IMAGE:latest" "$@"
```
netbrain: I feel that this is less reabable now
```suggestion
$CONTAINER_TOOL run "${GENERAL_FLAGS[@]}" \
    "${VGA_DEVICE_FLAG[@]}" \
    "${PODMAN_FLAGS[@]}" \
    "$IMAGE:latest" \
    "$@"
```

maybe we can tweak or disable some rules so that it doesn't warn about multiple lines like this?

entrypoint.sh
```diff
@@ -2,13 +2,13 @@
 set -e
 set -x
 
-# Check whetehr we are running in Docker/ Podman
-# Docker has the file /.dockerenv 
+# Check whether we are running in Docker/ Podman
+# Docker has the file /.dockerenv
 # Podman exposes itself in /run/.containerenv
-if ! ( [[ "$CONTAINER" == "docker" ]] || [[ "$CONTAINER" == "podman" ]] ); then
+if [[ ! "$CONTAINER" == "docker" ]] && [[ ! "$CONTAINER" == "podman" ]]; then
```
netbrain:
```suggestion
if [[ "$CONTAINER" != "docker" ]] && [[ "$CONTAINER" != "podman" ]]; then
```

zwift.sh
```diff
@@ -406,6 +402,6 @@ if [ $RC -ne 0 ]; then
 fi
 
 # Allow container to connect to X, has to be set for different UID
-if [ -n "$CONTAINER" ] && [ -x "$(command -v xhost)" ] && [ -z $WINE_EXPERIMENTAL_WAYLAND ]; then
-    xhost +local:$($CONTAINER_TOOL inspect --format='{{ .Config.Hostname  }}' $CONTAINER)
+if [ -n "$CONTAINER" ] && [ -x "$(command -v xhost)" ] && [ -z "$WINE_EXPERIMENTAL_WAYLAND" ]; then
+    xhost +local:"$($CONTAINER_TOOL inspect --format='{{ .Config.Hostname  }}' "$CONTAINER")"
```
netbrain: this inner quote of the "$CONTAINER" surely doesn't work?
```suggestion
    xhost +local:"$($CONTAINER_TOOL inspect --format='{{ .Config.Hostname }}' $CONTAINER)"
```

bin/build-image.sh
glennvl: That was actually me who put that on a single line since I had to modify that line to add quotes and figured it was a short line anyway. The linter does not complain about line breaks. But since you prefer to see that command in multiple lines, I split it again in 6bd0c2e.

entrypoint.sh
glennvl: Changed to use `!=` in 821b515.

zwift.sh
glennvl: No, that does work because the nested quotes are inside a `$(...)` block. I tested it myself because I was suspicious too. The linter wants the quotes there. I did remove the extra space in 6eb744e.

glennvl:
I tested what I can with podman. So if someone else can also test with docker that would be great.

netbrain:
I think we'll just assume it works with docker, and fix any issues that pops up. :crossed_fingers:

PR #237, opened 2025-10-21
Lint and spellcheck README.md

glennvl:
Since we're on a roll linting everything, we might as well check the `README.md` file too.

This PR adds two checks:

* Validate the `README.md` markdown using [markdownlint](https://github.com/DavidAnson/markdownlint/)
* Check the spelling in the `README.md` file using [cspell](https://cspell.org/)

Unlike linting the nix flake, dockerfile and bash scripts, there is some configuration required for markdownlint and cspell. So configuration files were added and some choices had to be made.

* **.markdownlint-cli2.jsonc**: This is the configuration file for the markdown lint cli tool (used by the github action and vscode extension). In this file we configure which files are to be checked. In our case only `README.md`.
* **.markdownlint.json**: The configuration file for markdownlint itself. Where we can configure rules (enable/disable and options). By default it doesn't allow inline html, so I set that one to enabled. It also does not allow duplicate headings by default, which I usually set to only don't allow duplicate headings for siblings. This is not a rule that is currently triggered in the README file, but I think is a good one to have in general for markdown files. Another rule that is relevant is the check for line lengths. I've disabled it for code blocks and tables, but set it to 128 (which is the value I usually use) for all other lines. If another value is preferred, just let me know and I'll update it.
* **.cspell.json**: CSpell configuration file. Configure which language to use, which files to check, add a list of extra words that should be considered correct. And also some more advanced configuration such as exclude regexes. When running a spell checker on a file that contains code it is always a challenge to avoid too many false positives. So I added a regular expression to exclude (both inline and fenced) code blocks. The disadvantage is that it won't find spelling errors in code blocks, but the advantage is it will result in a lot less false positives and allow using inline code blocks to be used for words that are not valid english but we don't want to add to the dictionary such as names of tools and parameters.


.markdownlint.json
```diff
@@ -0,0 +1,13 @@
+{
+    "default": true,
+    "line-length": {
+        "line_length": 128,
```
glennvl: Maybe setting the line length to 108 would be better? That would match what github uses as width when it renders the readme file.

netbrain: when it renders it where?

netbrain:
are all these extra files really necessary? im looking at the i.e like [.markdownlint-cli2.jsonc](https://github.com/netbrain/zwift/pull/237/files#diff-4929acf5ce7500ed4d3f00992a1c918e702427b7857fed53152e48fd58c2ae9c) and the two other dot prefixed files?

Or can we at the very least hide them somewhere else, so we don't have so many turds in the root folder.

.markdownlint.json
glennvl: On the home page of the git project. So when visiting https://github.com/netbrain/zwift.

glennvl:
> are all these extra files really necessary? im looking at the i.e like [.markdownlint-cli2.jsonc](https://github.com/netbrain/zwift/pull/237/files#diff-4929acf5ce7500ed4d3f00992a1c918e702427b7857fed53152e48fd58c2ae9c) and the two other dot prefixed files?
> 
> Or can we at the very least hide them somewhere else, so we don't have so many turds in the root folder.

Yes they are required. It is probably possible to move them somewhere else, but that would make it a bit harder to run the linters locally. I plan to create a devcontainer too so all linters can be easily run locally (automatically in e.g. vscode). In most projects linter config files are in the root and the project files themselves are in subdirectories such as src.

That said. Yes we need them, yes it's probably possible to move them somewhere else. So where do you suggest we move them?

glennvl:
> > are all these extra files really necessary? im looking at the i.e like [.markdownlint-cli2.jsonc](https://github.com/netbrain/zwift/pull/237/files#diff-4929acf5ce7500ed4d3f00992a1c918e702427b7857fed53152e48fd58c2ae9c) and the two other dot prefixed files?
> > Or can we at the very least hide them somewhere else, so we don't have so many turds in the root folder.
> 
> Yes they are required. It is probably possible to move them somewhere else, but that would make it a bit harder to run the linters locally. I plan to create a devcontainer too so all linters can be easily run locally (automatically in e.g. vscode). In most projects linter config files are in the root and the project files themselves are in subdirectories such as src.
> 
> That said. Yes we need them, yes it's probably possible to move them somewhere else. So where do you suggest we move them?

Moving them to a `config` subdirectory proved to be harder than expected. I can get it to work with the action, but then I can no longer locally run the linting tools (they expect the config files to be in the project root), which is also important. It turned out that it is possible to merge the two `markdownlint` config files into one file, so two files remain instead of three.

.github/workflows/lint_readme.yaml
```diff
@@ -0,0 +1,29 @@
...
+          globs: ""
+  spellcheck-readme:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository code
+        uses: actions/checkout@v5
+      - name: Spellcheck README.md
+        uses: streetsidesoftware/cspell-action@v7
+        with:
+          config: '.cspell.json'
```
netbrain: I'd prefer if this also is moved into .github/resources or something like this. alternatively .config/ as described in https://cspell.org/docs/Configuration

but in the latter case we should probably standardize on using .config for config related files for workflows and ide's alike.

.github/workflows/lint_readme.yaml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  validate-readme-markdown:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository code
+        uses: actions/checkout@v5
+      - name: Check README.md file for errors
+        uses: DavidAnson/markdownlint-cli2-action@v20
+        with:
+          config: '.markdownlint-cli2.jsonc'
```
netbrain: Move this aswell, see other comment.

.github/workflows/lint_readme.yaml
glennvl: Moved to .config directory in 5435655.

.github/workflows/lint_readme.yaml
glennvl: Moved to .config directory in 5435655.

.markdownlint.json
glennvl: @netbrain What do you want to do with the line length? Keep it at 128 or change it to something else?

netbrain: 128 is probably a good start. We'll simply change it later if we find something better.

PR #239, opened 2025-10-22
Lint Dockerfile and squash.Dockerfile (best practices)

glennvl:
The docker linter used in #233 only checks for errors in the dockerfiles. In this merge request, which builds further on #233, an additional linter [hadolint](https://github.com/hadolint/hadolint) is added that also checks for best practices.

Changes made

* Update rust image
* Consolidate `RUN` instructions
* Cleanup package manager after installing and avoid installing unnecessary packages
* Remove obsolete installation of `curl` in the zwift image (it is already installed in `wine-base`)

This linter also requires a config file since there's a few rules that we should (probably) disable:

* `DL3007` - Pin to a specific tag (don't use latest). `squash.Dockerfile` uses `netbrain/zwift:latest`.
* `DL3008` - Pin package versions. I don't think we want to pin all package versions in `apt-get`. 
* `DL3059` - Use only one `RUN` instruction per image. We use multiple, we could merge them into one at the cost of readability.

It is possible to avoid the config file if we disable `DL3007` and `DL3008` with an inline suppress in the affected dockerfile. And if we either merge the `RUN` statements into a single one per image or if we again suppress the rule in the dockerfile itself. In my opinion merging the `RUN` statements for the `build-runfromprocess` image would be fine, but in the `wine-base` image it would come at the cost of readability.

I would personally lean to keep the `.hadolint.yaml` config file, but if @netbrain prefers to not have the config file in the project root I'm happy to replace it with suppresses in the dockerfiles.

With this additional check we have covered linting (warnings/error and best practices) for all relevant files. The nix flake, the scripts, and the docker files.

**Note**: Depends on #233 to be merged first.


glennvl:
> move .hadolint.yaml to .github/workflows or maybe .github/resources
> 
> and refactor lint_dockerfile.yaml to set config to correct path, and i think we can merge this.

Moved to .config directory in 77244fa.

PR #242, opened 2025-10-31
Fix shebang for NixOs

avandecreme:
I made two possible fix (see the two commits):
- revert the change from #232 
- patch the shebang in the flake

Both work for me but the second one has the disadvantage that the script will always report as being out of date.
This could be workaround by ignoring the first line of the script in the sha256sum though.
Let me know what you prefer.

There are other scripts with the `#!/bin/bash` shebang but from what I understand they are run either by the CI or from within a container, so should be fine.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -73,6 +73,7 @@
 
         installPhase = ''
           runHook preInstall
+          patchShebangs .
           install -Dm755 ${nixosRun} -T $out/bin/${pname}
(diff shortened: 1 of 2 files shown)


netbrain:
Thanks for the contrib! Haven't run zwift myself for a while, so haven't detected the issue myself.

avandecreme:
I'm afraid my fix with patchShebangs doesn't work. Apologies about that :S

I'm unable to figure out the correct incantation. Would you like a rollback the first commit solution?

PR #244, opened 2025-11-03
Fix shebang for NixOs using /usr/bin/env bash 

netbrain:
Reverts netbrain/zwift#242


netbrain:
@avandecreme FYI

glennvl:
@netbrain Would it then not be better to change the shebang in all scripts? I'm a man of consistency. If they're all the same it would also set an example of what to use if a new script, that may also need to run in nix, is added.

netbrain:
Yes that's a good approach. I simply did what was strictly required to get it working again for nix.

PR #250, opened 2026-01-03
Debian forky with latest wine

glennvl:
* Update to debian:forky-slim and latest wine version.
* Allow container to access bluetooth device.

### DIFF
--- a/bin/build-image.sh
+++ b/bin/build-image.sh
@@ -41,9 +41,13 @@ GENERAL_FLAGS=(
 
 # Check for proprietary nvidia driver and set correct device to use
 if [[ -f "/proc/driver/nvidia/version" ]]; then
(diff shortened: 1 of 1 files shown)


glennvl:
```console
+ wine ZwiftSetup.exe /SP- /VERYSILENT /SUPPRESSMSGBOXES /NORESTART /NOCANCEL
X Error of failed request:  BadValue (integer parameter out of range for operation)
  Major opcode of failed request:  130 (MIT-SHM)
  Minor opcode of failed request:  3 (X_ShmPutImage)
  Value in failed request:  0x300
  Serial number of failed request:  319
  Current serial number in output stream:  319
```

Zwift installer currently crashes with glx. Zwift works with glx after letting the installer fail to completion with egl.

PR #253, opened 2026-01-17
Add 'Troubleshooting' paragraph

mndckr:
and explain how to enable multicast f…or wifi trainers on distros using firewalld.
 
 Disclaimer: Used AI for formatting and formulating the paragraph, but the content to put there was specified and tested by me .

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,6 +393,70 @@ environment variables in camelCase:
 }
 ```
 
+## Troubleshooting
+
+<details>
+<summary><h3>My WiFi-capable trainer / Zwift Companion App is not detected</h3></summary>
+
+If you have issues with device detection over WiFi/network, the issue may be related to your system's firewall.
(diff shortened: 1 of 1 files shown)


netbrain:
Thanks for putting this together — this is genuinely useful troubleshooting info for firewalld users!

That said, I'm wondering if this might fit better as a **pinned Discussion** rather than in the README. The README is already fairly long, and I'd like to keep it focused on setup and configuration. Troubleshooting content tends to grow over time as more edge cases surface, and Discussions would allow the community to add their own solutions and follow-ups.

Would you be open to posting this as a Q&A Discussion instead? I can pin it so it stays visible. We could then add a brief "Having issues?" note in the README linking to Discussions.

Let me know what you think!

mndckr:
I think either is fine. I'm on the same page regarding the readme length, which is why I made the section to be collapsed by default when rendered.

mndckr:
Fixed the CI

PR #257, opened 2026-01-24
Update ZWIFT_PASSWORD workflow to also support docker

glennvl:
As noted in discussion https://github.com/netbrain/zwift/discussions/256, automatically logging into zwift does not work with docker as its secret tool only works in swarm mode. This PR updates the workflow for ZWIFT_PASSWORD such that it works with both docker (use an environment variable) and with podman (use a secret).

```mermaid
flowchart LR
    A[[Start]] -->  C{ZWIFT_PASSWORD set?}
    C -->|No| D{container tool is podman and container tool secret exists?}
    D -->|Yes| E[SECRET_FOUND=True]
    D -->|No| F[Try to get ZWIFT_PASSWORD from linux secret-tool]
    E --> G( )
    F --> G
    C -->|Yes| G
    G --> L{ZWIFT_PASSWORD set?}
    L -->|Yes| H[PASSWORD_SET=True]
    H --> I[If podman, try to create/update container tool secret]
    I --> J{Secret created/updated successfully?}
    J -->|Yes| K[SECRET_FOUND=True]
    J -->|No| M[[End]]
   L -->|No| M
   K --> M
```

```mermaid
flowchart LR
    A[[Start]] -->  B{SECRET_FOUND ?}
    B -->|Yes| C[Use --secret for password]
    B -->|No| D{PASSWORD_SET?}
    D -->|Yes| E[Use -e for password]
    D --> |No| F[[End]]
    E --> F
    C --> F
```

### DIFF
--- a/zwift.sh
+++ b/zwift.sh
@@ -152,9 +152,13 @@ if [ -n "$ZWIFT_USERNAME" ]; then
     elif [[ $HAS_PLAINTEXT_PASSWORD -eq "1" ]]; then
         ZWIFT_PASSWORD_SECRET="-e ZWIFT_PASSWORD=$ZWIFT_PASSWORD"
     else
-        echo "No password found for $ZWIFT_USERNAME..."
-        echo "To avoid manually entering your zwift password every time, either start zwift with \`ZWIFT_PASSWORD=\"hunter2\" zwift\` or store it in the secret-store"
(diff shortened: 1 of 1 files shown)


netbrain:
Great stuff! I'll look into it later

keesjan42:
I have tested this change on my PC and for me it works when I save my password in secret-tool beforehand.
Very nice and very fast correction!
Of course I have not tested all other cases.

netbrain:
LGTM

PR #258, opened 2026-01-25
Fancy messages

glennvl:
I was playing around a bit with colors in bash. So I gave it a go to add some more feedback of what the zwift.sh script is doing and made the messages a bit fancier. Up to you all if you want this or not. :smile:

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -43,6 +43,11 @@ msgbox() {
     esac
 }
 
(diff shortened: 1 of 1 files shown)


netbrain:
I don't mind me some colorful panash! :)

glennvl:
> I don't mind me some colorful panash! :)

I also updated the install script :innocent:

Issue #254, opened 2026-01-17
NixOS module not picking up configuration options


szaffarano:
**Describe the bug**

When using the NixOS module, it ignores all the module's config options.

**To Reproduce**
1. Install `zwift` following the readme's instructions and customizing at least one option, e.g.,  the username
    ```nix
        programs.zwift = {
          enable = enabled;
          zwiftUsername = "my-user@domain.com";
        };
    ``` 
2. Configure your zwift creds as stated in the readme: `secret-tool store --label "Zwift password for my-user@domain.com" application zwift username my-user@domain.com`
3. Launch zwift from a terminal

Zwift should pick the configured user up and pass the creds to the container, i.e.
```bash
❯ zwift 
Looking up Zwift password for my-user@domain.com...
Passing password and username to zwift container.
```

**Screenshots**
N/A

**Additional context**

The problem is that the module does not export the defined environment variables, so the downstream script, `zwift-nixos.sh`, cannot read them. I think that the fix would be as simple as prepending the export keyword, e.g.

```nix
nixosRun = pkgs.writeShellScript "zwift-nixos.sh" ''
  export IMAGE=${image}
  export VERSION=${tag}
  export DONT_CHECK=${dontCheck}
  export DONT_PULL=${dontPull}
  export DRYRUN=${dryRun}
  export INTERACTIVE=${interactive}
  export CONTAINER_TOOL=${containerTool}
  export CONTAINER_EXTRA_ARGS=${containerExtraArgs}
  export ZWIFT_USERNAME=${zwiftUsername}
  export ZWIFT_PASSWORD=${zwiftPassword}
  export ZWIFT_WORKOUT_DIR=${zwiftWorkoutDir}
  export ZWIFT_ACTIVITY_DIR=${zwiftActivityDir}
  export ZWIFT_LOG_DIR=${zwiftLogDir}
  export ZWIFT_SCREENSHOTS_DIR=${zwiftScreenshotsDir}
  export ZWIFT_OVERRIDE_GRAPHICS=${zwiftOverrideGraphics}
  export ZWIFT_OVERRIDE_RESOLUTION=${zwiftOverrideResolution}
  export ZWIFT_FG=${zwiftFg}
  export ZWIFT_NO_GAMEMODE=${zwiftNoGameMode}
  export WINE_EXPERIMENTAL_WAYLAND=${wineExperimentalWayland}
  export NETWORKING=${networking}
  export ZWIFT_UID=${zwiftUid}
[...]

PR #260, opened 2026-01-26
fix(nix): Improve NixOS module configuration and docs

szaffarano:
- Define the `containerTool` as an enum and enable the proper virtualisation (docker or podman) depending on this parameter.
- Export environment variables in script
- Add firewall configuration snippet for Wahoo / Zwift click devices.

This PR fixes https://github.com/netbrain/zwift/issues/254

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,7 +393,7 @@ environment variables in camelCase:
 }
 ```
 
-### Firewal configuration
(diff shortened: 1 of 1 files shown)


netbrain:
Looks correct to me, but i'm no expert. Trusting you on this! :+1:

Issue #261, opened 2026-01-27
Passwords with spaces does not work with secret-tool


msouthcott:
**Describe the bug**
I had a password with spaces and I set ENV:$DRYRUN to see what the actual command was that zwift wsa running.  When I removed my password zwift worked.  But when a password with spaces is used it fails.  I can't really post the failure because it uses part of my password as a string in the failure.  Here is the output from DRYRUN.

docker run --rm --network bridge --name zwift-matt --hostname thinkpad -e DISPLAY=:0 -e ZWIFT_UID=1000 -e ZWIFT_GID=1000 -e PULSE_SERVER=/run/user/1000/pulse/native -e CONTAINER=docker -v zwift-matt:/home/user/.wine/drive_c/users/user/Documents/Zwift -v /run/user/1000/pulse:/run/user/1000/pulse --security-opt label=type:container_runtime_t -d --env-file /home/matt/.config/zwift/config -e ZWIFT_USERNAME=someone@gmail.com -e ZWIFT_PASSWORD=something space --device=/dev/dri:/dev/dri -e DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus -v /run/user/1000/bus:/run/user/1000/bus -e XAUTHORITY=/run/user/1000/.mutter-Xwaylandauth.KSNHI3 -v /tmp/.X11-unix:/tmp/.X11-unix -v /run/user/1000/.mutter-Xwaylandauth.KSNHI3:/run/user/1000/.mutter-Xwaylandauth.KSNHI3 docker.io/netbrain/zwift:latest 

OUTPUT from above command
✓] Password found for Matt.southcott@gmail.com
[*] Could not create secret for password, using environment variable instead
[*] Checking for updated zwift.sh
[✓] You are running the latest zwift.sh 👏
[*] Checking for updated container image
latest: Pulling from netbrain/zwift
Digest: sha256:965247a9eb2eab85d459b50f47b783e2ccb085967e5e1ee566487d329d896e1b
Status: Image is up to date for netbrain/zwift:latest
docker.io/netbrain/zwift:latest
[✓] Container image is up to date
Unable to find image 'space:latest' locally
docker: Error response from daemon: pull access denied for or, repository does not exist or may require 'docker login': denied: requested access to the resource is denied

**To Reproduce**
1. have a password with spaces
2. setup secret-tool then run ./zwift
3. error varies depending on password.  It takes first word as password second word is parsed incorrectly.

**Expected behavior**
[...]

Issue #259, opened 2026-01-26
New version of script doesn't load credentials from config file


OSLDNKKfEUVTABDLiNgk:
**Describe the bug**
Using the latest version of the script, the Zwift launcher does not automatically log in. Seems to ignore credentials stored in 

> /home/user/.config/zwift/config

**To Reproduce**
1. Update to latest version of script
2. Store username and password in config file
3. Launch Zwift

**Expected behavior**
Zwift would log in automatically.

**Screenshots**

<img width="2256" height="1504" alt="Image" src="https://github.com/user-attachments/assets/851492c1-11be-48dd-a390-8d55f8f3e371" />

**Additional context**
The internet connection is fine. I am able to log in and use Zwift like normal once I enter credentials. I don't use the podman secret store. Running on Bazzite: 

> user@bazzite:/var/home/user$ zwift
Looking up Zwift password for user@email.com...
Stored password in podman secret store
You are running latest zwift.sh 👏
Trying to pull docker.io/netbrain/zwift:latest...
Getting image source signatures
Copying blob 79662a5f39ae skipped: already exists  
Copying blob 7b4a7bfaaa47 skipped: already exists  
Copying blob 0a5b42a4aa90 skipped: already exists  
Copying blob 57444c22cb17 skipped: already exists  
Copying blob c2fab3b65828 skipped: already exists  
Copying blob c1432599aa04 skipped: already exists  
Copying blob af2c9c97d393 skipped: already exists  
Copying blob 999bee9fdfe5 skipped: already exists  
Copying blob 0256d6a7b130 skipped: already exists  
Copying blob fa12bcdc441e skipped: already exists  
Copying blob 0fbcba8041d5 skipped: already exists  
Copying blob 4f4fb700ef54 skipped: already exists  
Copying blob f73286002ef4 skipped: already exists  
Copying blob 4f4fb700ef54 skipped: already exists  
Copying blob 748f4b314aeb skipped: already exists  
Copying blob f0f7e48b3c96 skipped: already exists  
Copying blob de6068f869e3 skipped: already exists  
Copying blob 4f4fb700ef54 skipped: already exists  
Copying blob 4f4fb700ef54 skipped: already exists  
Copying blob aa6becb46061 skipped: already exists  
Copying blob 4f4fb700ef54 skipped: already exists  
Copying blob ad3f17dccd3b skipped: already exists  
Copying blob fbf52c050a80 skipped: already exists  
Copying blob 99c48c102915 skipped: already exists  
[...]

Issue #190, opened 2025-02-02
Keyring not being accessed by Docker Container


Guacam-Ole:
**Describe the bug**
I try to use the secretservice to store the password as suggested. But when starting the docker container it cannot be found. Using the pw in the config works

**To Reproduce**
Store the password in secretservice, start zwift

**Additional context**
When starting the container the following message appears:

```
Password not found in the SecretService keyring, you can store it with `secret-tool store --label "Zwift password for [myemailadress]" application zwift username [myemailadress]`
```

So I did exactly like suggested in the message and still receive that error. Maybe docker cannot access the secret store?

netbrain:
@Jmennius you have any clue?

Jmennius:
@Guacam-Ole can you try running manually `secret-tool lookup application zwift username <username>` (substitute your configured username here), what is the output (does it show the password or is there an error? you don't have to show us the password of course)?

Jmennius:
relates to #178

GrahamJohnson:
The option to use secret-tool to store the Zwift password does not work at all on docker.  It appears to have been written and tested with podman in mind, and not tested at all with docker. 

1. Docker seems to require 'docker swarm init' for secrets to be created (and, presumably, used), which is not set up by zwift.sh.
2. Docker secret names cannot contain @ (or +) characters, so Zwift usernames (i.e. email addresses) cannot be used as-is as docker secret names.  @ and + would have to be escaped.
3. The 'docker create secret ' command (when it omits the podman-only flag to replace the secret) passes two spaces in a row, which seems to break the command.  This can be fixed by putting the extra space in the optional flag string and removing it from before the secret name.
4. Docker will give an error if you try to create a secret that already exists, so if you change your password in secret-tool the change will not make it to docker.  Logic would need to be added to always try to remove the existing secret from docker ('docker secret rm') before trying to add it.
5. 'docker secret exists' is not a valid command.  Need to use 'docker secret ls' instead.
6. Docker does not have a --secret argument, so the way zwift.sh passes that to try to copy the secret to an environment variable does not work as written.  I do not know what the docker equivalent of this would be.

I made changes (see below) to my /usr/local/bin/zwift  to correct some of this, but gave up when I encountered  the last issue.

```diff
125c125,127
<     PASSWORD_SECRET_NAME="zwift-password-$ZWIFT_USERNAME"
---
[...]

PR #262, opened 2026-01-27
Generate temporary env file

glennvl:
## Summary

Use an env file instead of passing environment variables on the command line. This is a workaround for environment variables that have special characters such as backslashes, spaces, ... in their value. Especially revelant for passwords since those tend to include special characters. The advantage of using an env file is that environment variables can be written as `KEY=VALUE` without requiring any escape characters, quotes, ...

## Implementation

* Create an array variable in which environment variables can be appended.
* Create a temporary file that will be used as env file.
* Add environment variables to the array instead of adding `-e VARIABLE=VALUE`.
* Before starting the container, dump the array of environment variables to the temporary env file.
* When the script exits, automatically remove the env file (using an exit trap). When the container has started the file is no longer needed and can be safely removed (no need to wait for the container to exit).

The temporary file is created using `ENV_FILE=$(mktemp -q /tmp/zwift-container.env.XXXXXXXXXX)`. Which results in a file with permissions `600` (rw for the owner, no access for everyone else) and owner and group being the user that invoked the script. This should be as safe as we can get to temporarily store the zwift password in plain text format (only relevant when using docker, when using podman the secret store is used instead of plain text password).

## Debugging

Since we now use an env file for environment variables instead of passing them with `-e` on the commandline, the environment variables are not present anymore when looking at the commandline with for example `DRYRUN=1`.

* When using `DRYRUN=1`, the list of environment variables will also be printed in addition to the commandline. So all the information is still there. Note that if a password is present in the list of environment variables, it will be printed as `ZWIFT_PASSWORD=REDACTED` instead of showing the actual password.
* Security consideration: When using `DRYRUN=1`, the environment variables are only printed to the screen and never actually written to the env file. Since the container never actually starts, it's not necessary to actually create the file, so we can avoid exposing the password in plain text.
* Alternatively, `INTERACTIVE=1` can be used to launch the container in interactive mode. In the container the environment variables can be inspected with the `env` command. This will print a list of all environment variables (including the plain text password, for docker and podman).

## Sourcing config files and passing them as env files

The config files used to be both sourced into the zwift script and passed as env files to the container. This PR no longer passes the config files as env files to the container. Because files that are supposed to be sourced and env files have different syntax requirements.

* Files that are sourced
  * Are treated as bash scripts, they can contain both code and variable assignments
  * The variable assignments need to be properly escaped or single-quoted
* Environment files
  * Can only contain `key=value` entries and lines starting with a `#` are treated as comments
  * Values must not be quoted or escaped, all characters are literal
* Examples that show difference in behaviour
  * `KEY=my value`
     * source: `KEY` is an array of two values `my` and `value`
     * env file: `KEY` has the value `my value`
  * `KEY=my value \`
    * source: syntax error trying to parse file
    * env file: `KEY` has  the value `my value \`  
  * `KEY='my value \'`  
    * source: `KEY` has the value `my value \`
    * env file: `KEY` has the value `'my value \'`
  * `KEY = value # something`
    * source: `KEY` has the value `value` (`somthing` is treated as a comment)
    * env file: `KEY` has the value `value # something`

Both sourcing and using the files as env files could easily result in errors and different values for environment variables in the zwift script and the container.

## Discussion

Since the config files are no longer passed as env files to the container, the user can no longer inject their own environment variables in the container.

* Is it necessary for the user to be able to do so?
* If yes, we can easily implement a new feature to allow the user to supply their own env file with environment variables to be passed to the container.

   ```bash
   ADDITIONAL_ENV_FILES=()   

   load_env_file() {
       EXTRA_ENV_FILE="$1"
       msgbox info "Looking for env file $EXTRA_ENV_FILE"
       if [[ -f "$EXTRA_ENV_FILE" ]]; then
           msgbox ok "Using $ENV_FILE"
           ADDITIONAL_ENV_FILES+=(--env-file "$EXTRA_ENV_FILE")
       fi
   }

   load_env_file "$HOME/.config/zwift/container.env"
   load_env_file "$HOME/.config/zwift/$USER-container.env"

   # Add "${ADDITIONAL_ENV_FILES[@]}" to CMD.
   ```

## Related issues

* #261
* #259 
* #190

### DIFF
--- a/zwift.sh
+++ b/zwift.sh
@@ -451,7 +451,9 @@ if [[ -n "$DRYRUN" ]]; then
     msgbox ok "DRYRUN:"
     msgbox ok "environment variables ($ENV_FILE):"
     for env_var in "${ENVIRONMENT_VARIABLES[@]}"; do
-        msgbox ok "  ${env_var//ZWIFT_PASSWORD=*/ZWIFT_PASSWORD=REDACTED}"
+        env_var=${env_var//\\/\\\\}                                  # escape backslashes
+        env_var=${env_var//ZWIFT_PASSWORD=*/ZWIFT_PASSWORD=REDACTED} # redact password
+        msgbox ok "  $env_var"
     done
     msgbox ok "$CONTAINER_TOOL command:"
     msgbox ok "  $(printf '%q ' "${CMD[@]}")"


netbrain:
Great stuff, LGTM.

netbrain:
Not sure if we need a custom env functionality yet, but we'll see if anyone requests it for their use case.

kschat:
Hey all! I found this PR when I noticed Zwift was using the wrong timezone for my rides an update. For some reason the container uses UTC instead of my host machines timezone, and the easiest/fastest fix I found was manually setting the `TZ` environment variable in the config file.

> If yes, we can easily implement a new feature to allow the user to supply their own env file with environment variables to be passed to the container.

Good news! This isn't required, as there's an existing workaround. I was able to fix it with [`CONTAINER_EXTRA_ARGS` using `--env`](https://github.com/kschat/dotfiles/blob/master/zwift/config#L6). So no new feature is needed as long as `CONTAINER_EXTRA_ARGS` remains.

PR #270, opened 2026-02-02
Add missing dev tools to devcontainer

netbrain:
## Summary

Brings the devcontainer to parity with the Nix flake dev environment by adding missing tools:

- `nixfmt-rfc-style` (replaces `nixfmt` for consistency with flake)
- `shfmt` for bash formatting
- `actionlint` for GitHub Actions linting
- `yamllint` for YAML linting

Also adds VS Code extensions:
- `foxundermoon.shell-format` for shfmt integration
- `redhat.vscode-yaml` for YAML support

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -262,7 +262,7 @@
 
           # Nix
           nil
-          nixfmt-rfc-style
+          nixfmt
 
           # Docker
           hadolint


netbrain:
@glennvl mind testing this to see if this works for you?

.devcontainer/devcontainer.json
```diff
@@ -9,7 +9,7 @@
         },
         "ghcr.io/devcontainers/features/nix": {
             "version": "latest",
-            "packages": "nil,nixfmt"
+            "packages": "nil,nixfmt-rfc-style,shfmt,actionlint,yamllint"
```
glennvl: Nixfmt has been updated to point to nixfmt-rfc-style instead of nixfmt-classic (https://github.com/NixOS/nixpkgs/issues/425583). In the latest version of nixpkgs (which is used by the nix dev container feature) the old nixfmt-rfc-style package does not exist anymore. I think the issue is that the flake.lock file in the repository points to an old version of nixpkgs before those changes were made (nixfmt-rfc-style exists and nixfmt still points to classic?). So we should probably regenerate the lock file and use nixfmt from the latest nixpkgs?

.devcontainer/devcontainer.json
```diff
@@ -45,7 +47,7 @@
                     "serverSettings": {
                         "formatting": {
                             "command": [
-                                "nixfmt"
+                                "nixfmt-rfc-style"
```
glennvl: I think the command is always nixfmt, regardless of whether nixfmt, nixfmt-classic or nixfmt-rfc-style is installed.

.devcontainer/devcontainer.json
```diff
@@ -9,7 +9,7 @@
         },
         "ghcr.io/devcontainers/features/nix": {
             "version": "latest",
-            "packages": "nil,nixfmt"
+            "packages": "nil,nixfmt-rfc-style,shfmt,actionlint,yamllint"
         },
```
glennvl, line 17:
```json
        "ghcr.io/devcontainers/features/ruby": {
            "version": "latest"
        },
```

Ruby should also be installed for the jekyll docs.

Note that installing ruby breaks the path. So also add the following to make sure nix is installed after ruby and the path remains intact for everyone.

```json
    "overrideFeatureInstallOrder": [
        "ghcr.io/devcontainers/features/docker-in-docker",
        "ghcr.io/devcontainers/features/ruby",
        "ghcr.io/devcontainers/features/nix",
        "ghcr.io/devcontainers-extra/features/shellcheck",
        "ghcr.io/dhoeric/features/hadolint"
    ],
```

.devcontainer/devcontainer.json
glennvl:
```json
"postCreateCommand": "if ! docker info > /dev/null 2>&1; then sudo update-alternatives --set iptables /usr/sbin/iptables-nft; fi; gem update --system && gem install bundler",
```

Install ruby bundler as a gem in the post create command, unless you have a better solution 😃

.devcontainer/devcontainer.json
```diff
@@ -9,7 +9,7 @@
...
         "ghcr.io/devcontainers/features/nix": {
             "version": "latest",
```
glennvl, line 14: Maybe add `"extraNixConfig": "experimental-features = nix-command flakes"`? Doesn't hurt to have nix commands available in the vscode terminal.

glennvl:
@netbrain Looks good to me now! Just one remaining question, should the nix flake also be updated to `nixfmt` instead of `nixfmt-rfc-style` now that the lock file is updated?

netbrain:
Yes definitely, good catch! Updated in 37b47e1.

glennvl:
> Yes definitely, good catch! Updated in [37b47e1](https://github.com/netbrain/zwift/commit/37b47e128462134f2ab91070ab359443a29155f4).

Should be good to go to merge this once the github action is complete then :+1: 

(I'll keep an eye on the nix devcontainer feature if there's any progress on getting it to play nice with vscode plugins so we can eventually use the nix flake. But I suspect that won't be possible without changes to vscode itself.)

PR #271, opened 2026-02-02
Use Nix flake for devcontainer tools

netbrain:
## Summary

Alternative to PR #270. Instead of duplicating tool installations, this approach
uses the Nix flake as the single source of truth for dev tools.

Changes:
- Adds direnv feature and VS Code extension
- Removes individual tool installations (shellcheck, hadolint features)
- Uses `direnv allow` on container creation to activate the flake
- Tools are provided by `flake.nix` via `.envrc`

Benefits:
- Single source of truth for dev tools
- No tool version drift between flake and devcontainer
- Adding tools to the flake automatically makes them available in devcontainer

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,12 +9,9 @@
         },
         "ghcr.io/devcontainers/features/nix": {
             "version": "latest",
-            "packages": "nil,nixfmt"
+            "extraNixConfig": "experimental-features = nix-command flakes"
         },
-        "ghcr.io/devcontainers-extra/features/shellcheck": {
-            "version": "latest"
-        },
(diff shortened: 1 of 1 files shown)


netbrain:
@glennvl can you also review this as an alternative?

glennvl:
> @glennvl can you also review this as an alternative?

This would be the way to go. But I've been trying to get it to work all afternoon now, and it just won't cooperate. The environment is loaded at the wrong time, so vscode does not pick it up. Extensions don't find the tools installed through the nix flake. I can get it to activate direnv in the built-in console, but only if I disable the vscode direnv extension (it causes an infinite reload loop). I don't think getting this in a working state is going to happen to be honest.

netbrain:
Man, thats a shame. But ok. Let's move on, and just use #270 instead.

PR #272, opened 2026-02-03
Update project structure

glennvl:
## Disclaimer

This is another one of my proposals where I make significant changes to the project. :see_no_evil: 

## Summary

Since the project is becoming a bit more mature with regards to completeness (documentation, contributing, ...), the root directory of the project is also growing. Which of course is a normal thing.

This PR updates the project structure to a more mature structure that separates source code from metadata.

## Implementation

Follow the often used pitchfork layout by creating a src and install directory:

* The install directory replaces the assets directory. It contains both the assets and the install script
* The src directory contains the actual source code of the project (dockerfiles, scripts, and other related files)

Other changes to the project structure:

* Move config files to the top level directory where they are expected by the tools, thus making it easier for contributors and ide plugins to invoke tools correctly
* Convert all config files to yaml for consistency (this was possible since all tools also support yaml)

## Discussion

The main zwift script automatically updates itself by comparing its checksum to the checksum of the zwift script in master. If the checksums are different, the install script is downloaded from master and invoked to install the latest zwift script.

* The zwift script no longer lives in the project root, it was moved to the src directory. When comparing checksums to see if an update is available, the zwift script is downloaded from master. When the user still has the previous zwift script that expects zwift.sh to live in master, curl will return a 404 code. This 404 code is then used to calculate the checksum, which differs from the local zwift.sh checksum. The script will correctly determine a new version is available. So the zwift script can be safely moved to the src directory without causing any issues.
* When the old script determined it needs to be updated, it will try to download the install script from master. This poses an issue since it no longer lives there, instead it was moved to the install directory. To work around this issue, a new very small install script was introduced in the project root. All this script does is redirect to the new install script by dowloading and invoking the install script from the install directory. It's not perfect, but I think its preferred over keeping the real install script in the project root between config files and readmes where it would be a bit out of place.

## :warning: TODO - Nix flake must be updated before merging

I'm not a nix user, so I wasn't sure what exactly to change in the nix flake. If someone could provide a patch file or directions on what to change, that would be great!

- [x] Move installation related files to install directory
- [x] Move source code to src directory
- [x] Move metadata/config files to the root directory
- [x] Update nix flake

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -88,7 +88,8 @@
           installPhase = ''
             runHook preInstall
             install -Dm755 ${nixosRun} -T $out/bin/${pname}
-            install -Dm644 $src/bin/zwift.svg -T $out/share/icons/hicolor/scalable/apps/zwift.svg
+            install -Dm644 $src/share/icons/hicolor/scalable/apps/zwift.svg \
+                    -T $out/share/icons/hicolor/scalable/apps/zwift.svg
             runHook postInstall
           '';
 


netbrain:
What I like:                                                                                                                                                                                
  - Configs at root, provides cleaner IDE setup
  - Grouping source files in src/                                                                                                                                                      
         
  What I dislike:
  - The redirect install.sh at root
  - That redirect script calls pkexec, then the real install script calls it again
  - flake.nix still references zwift.sh without the src/ prefix
  - Not sure the reorg is worth the breaking change for a project this size
  
  Maybe a solution is to have the install script at root aswell.

glennvl:
> What I dislike:
> 
>     * The redirect install.sh at root
> 
>     * That redirect script calls pkexec, then the real install script calls it again

I agree that it is not ideal. But keep in mind that the install script at the root of the project is **only** used if the zwift script is updated from the current version that expects the install script to be at the root. Once updated to the zwift script from this pr, the install script in the install directory is used, bypassing the install script at the root with redirect.

It should be possible to eventually remove the install script at the root altogether. When we expect all (most) users to have updated to the new version. Maybe in a few months?

> What I dislike:
> 
>     * flake.nix still references zwift.sh without the src/ prefix

Yup, still need to update the nix flake. But I got a bit confused seeing some paths in it prefixed by `$src` and some being relative paths. ~~I attempted to update it in cedface188cafc49ac58c4daaeda820287860468, could you check if that's correct? I did it wrong, didn't I? I shouldn't have touched the stuff in `pkgs.stdenv.mkDerivation`?~~ I think I now actually fixed it in c964e0275495b47b6471fa36dc57d9600d313130 :sweat_smile: .

glennvl:
@netbrain I just caught myself being dumb. The expected location for the install script is of course not the root directory, it used to live in bin, not root, so basically I broke stuff now.

* **Option 1**: Move the install.sh redirect script to bin/install.sh, keeping the redirect script and hopefully eventually deleting it.
* **Option 2**: If we remove the install script with redirect from the root directory and rename the install directory to bin, there's no breaking changes. The only disadvantage that our install directory would be called bin instead of install (not ideal, but something we can live with?). I went ahead and implemented this in d6f3f1cf8ab4e2c9499e7a6105268eced3b31d7f.

netbrain:
bin works for me! this ain't a c++ project! :) I think it looks good now, and seems ready to merge. Again, super good job! nice to have someone who can do some housekeeping and putting some makeup on this pig :rofl:

netbrain:
@glennvl you mind looking into why https://github.com/netbrain/zwift/actions/runs/21742306795/job/62720160961 failed after this merge? Actually, i think we're just missing a dir creation step. see #275

Issue #273, opened 2026-02-05
Cleanup previous images when an updated version is downloaded


danieldean:
## Describe the bug

When new images are downloaded by the script previous images are kept. Eventually this cause storage space to run out (has happened to me). The images are quite large at 10GB+.

## To Reproduce

Keep  starting Zwift using the script and every time it updates a new image is downloaded.

## Expected behavior

The previous images are cleaned. I personally see no reason to keep anything but latest?

## Screenshots

See below:

<img width="742" height="251" alt="Image" src="https://github.com/user-attachments/assets/9e105bb5-1a15-4a86-8446-ff70c0d006e9" />

## Fix?

My suggested fix for this with Docker is:

```docker image rm $(docker images --filter "before=netbrain/zwift:latest" --format '{{.ID}}')```

This deletes all up the latest image. I think the same can be done for Podman. If this is acceptable I will create a pull request?

**Comment by glennvl** [comment]
That's something that's been on the TODO list for a while, but noone has picked up so far. (See https://github.com/netbrain/zwift/issues/156, it talks both about removing images on docker hub and removing downloaded images.)

Both docker and podman support the `--filter` and `--format`, so yeah, the command should be the same for both.

```bash
$(docker images --filter "before=netbrain/zwift:latest" --format '{{.ID}}')
```

This lists **all** images before netbrain/zwift:latest, also unrelated images, so it would delete too much.

I haven't tested the command, but I'd think something like the following might yield better results?

* First list all netbrain/zwift images
* Then filter out latest

```bash
docker images netbrain/zwift --format "{{.ID}} {{.Tag}}" | awk '$2 != "latest" { print $1 }'
```

danieldean:
Quite right! I believe the correct command would be:

```docker image rm $(docker images --filter "reference=netbrain/zwift" --filter "before=netbrain/zwift:latest" --format '{{.ID}}')```

I looked at #156 and that the creator mentioned that only being about DockerHub.

PR #274, opened 2026-02-05
Add clean up of previous images

danieldean:
This is a contribution for #273. Let me know if there are any issues!

All tests passed apart from doc building. It works when running for me and cleans up as expected.

### DIFF
--- a/docs/configuration/options.md
+++ b/docs/configuration/options.md
@@ -16,8 +16,8 @@ These environment variables can be used to alter the execution of the zwift bash
 | `IMAGE`                     | `docker.io/netbrain/zwift` | The image to use                                                                                                                       |
 | `VERSION`                   | `latest`                   | The image version/tag to use                                                                                                           |
 | `DONT_CHECK`                |                            | If set, don't check for updated `zwift.sh`                                                                                             |
-| `DONT_PULL`                 |                            | If set, don't pull for new image version                                                                                               |
-| `DONT_CLEAN`                |                            | If set along with `DONT_PULL`, don't clean up previous image                                                                           |
+| `DONT_PULL`                 |                            | If set, don't pull for new image version (implies `DONT_CLEAN`)                                                                        |
+| `DONT_CLEAN`                |                            | If set, don't clean up previous image after pulling                                                                                    |
 | `DRYRUN`                    |                            | If set, print the full container run command and exit                                                                                  |
 | `INTERACTIVE`               |                            | If set, force `-it` and use `--entrypoint bash` for debugging                                                                          |
 | `CONTAINER_TOOL`            |                            | Defaults to podman if installed, else docker                                                                                           |


zwift.sh
```diff
@@ -297,6 +297,21 @@ if [[ ! $DONT_PULL ]]; then
     fi
 fi
 
+# Clean previous container images (if any)
+if [[ ! $DONT_CLEAN ]]; then
+    readarray -t images < <(IFS="/" read -r -a ref <<< "$IMAGE" ; $CONTAINER_TOOL images --filter "reference=${ref[1]}/${ref[2]}" --filter "before=${ref[1]}/${ref[2]}:$VERSION" --format '{{.ID}}')
```
netbrain:
```
 #!/usr/bin/env bash

  # The PR's parsing logic, extracted
  test_split() {
      local IMAGE="$1"
      local VERSION="latest"

      IFS="/" read -r -a ref <<< "$IMAGE"

      echo "IMAGE:    $IMAGE"
      echo "ref[0]:   ${ref[0]}"
      echo "ref[1]:   ${ref[1]}"
      echo "ref[2]:   ${ref[2]}"
      echo "filter:   reference=${ref[1]}/${ref[2]}"
      echo "before:   before=${ref[1]}/${ref[2]}:$VERSION"
      echo "---"
  }

  # WORKS
  test_split "docker.io/netbrain/zwift"

  # NO GO
  test_split "netbrain/zwift"

  # NO GO
  test_split "ghcr.io/org/sub/zwift"
```

Would yield 

```
     IMAGE:    docker.io/netbrain/zwift
     ref[0]:   docker.io
     ref[1]:   netbrain
     ref[2]:   zwift
     filter:   reference=netbrain/zwift
     before:   before=netbrain/zwift:latest
     ---
     IMAGE:    netbrain/zwift
     ref[0]:   netbrain
     ref[1]:   zwift
     ref[2]:
     filter:   reference=zwift/
     before:   before=zwift/:latest
     ---
     IMAGE:    ghcr.io/org/sub/zwift
     ref[0]:   ghcr.io
     ref[1]:   org
     ref[2]:   sub
     filter:   reference=org/sub
     before:   before=org/sub:latest
     ---
```

Maybe just use $IMAGE directly in the filter instead of splitting/reassembling it. E.g., --filter "reference=$IMAGE" and --filter "before=$IMAGE:$VERSION"?

zwift.sh
```diff
@@ -297,6 +297,21 @@ if [[ ! $DONT_PULL ]]; then
...
+    if [ "${#images[@]}" != "0" ]; then
+      msgbox info "Cleaning up previous container images"
+      if $CONTAINER_TOOL image rm "${images[@]}"; then
+          msgbox ok "Any previous container images have been deleted"
+      else
+          msgbox error "Failed to clean up previous container images"
+      fi
+    else
+      msgbox ok "No previous container images to clean up"
```
netbrain:
```suggestion
```

This just creates noise, so let's remove it.

zwift.sh
```diff
@@ -297,6 +297,21 @@ if [[ ! $DONT_PULL ]]; then
     fi
 fi
 
+# Clean previous container images (if any)
+if [[ ! $DONT_CLEAN ]]; then
+    readarray -t images < <(IFS="/" read -r -a ref <<< "$IMAGE" ; $CONTAINER_TOOL images --filter "reference=${ref[1]}/${ref[2]}" --filter "before=${ref[1]}/${ref[2]}:$VERSION" --format '{{.ID}}')
+    if [ "${#images[@]}" != "0" ]; then
+      msgbox info "Cleaning up previous container images"
+      if $CONTAINER_TOOL image rm "${images[@]}"; then
+          msgbox ok "Any previous container images have been deleted"
```
netbrain:
```suggestion
          msgbox ok "Previous container images have been deleted"
```

zwift.sh
```diff
@@ -297,6 +297,21 @@ if [[ ! $DONT_PULL ]]; then
     fi
 fi
 
+# Clean previous container images (if any)
+if [[ ! $DONT_CLEAN ]]; then
```
netbrain: We should probably only clean if DONT_CLEAN is not set AND DONT_PULL is not set?

zwift.sh
netbrain: Maybe an improvement would be to fork the "image rm" as to not block the rest of the script? This would hinder slowdown in zwift startup, especially critical if your late to your race ;)

netbrain:
Another thing, the dont_clean option should also be present in flake.nix

flake.nix
```diff
@@ -46,6 +47,7 @@
             ${pkgs.lib.optionalString (tag != "") "export VERSION=${tag}"}
             ${pkgs.lib.optionalString (dontCheck != "") "export DONT_CHECK=${dontCheck}"}
             ${pkgs.lib.optionalString (dontPull != "") "export DONT_PULL=${dontPull}"}
+            ${pkgs.lib.optionalString (dontClean != "") "export DONT_PULL=${dontClean}"}
```
netbrain:
```suggestion
            ${pkgs.lib.optionalString (dontClean != "") "export DONT_CLEAN=${dontClean}"}
```

glennvl:
@danieldean The zwift.sh script was moved to a src directory in #272, so you need to rebase your branch before these changes can be merged (`git rebase master`).

zwift.sh
danieldean: It now runs in a forked processed.

zwift.sh
danieldean: This was a particularly problematic part as the full `docker.io/netbrain/zwift` doesn't work which is why I did this. I will see if there is another way to resolve it as I already realised it wasn't very robust. I was trying to make sure the command was as strict as possible to avoid deleting any images we shouldn't.

danieldean:
> @danieldean The zwift.sh script was moved to a src directory in #272, so you need to rebase your branch before these changes can be merged (`git rebase master`).

Thanks and noted. At the moment I think the array splitting is a blocker so will resolve that first if possible then rebase.

zwift.sh
danieldean: So this is only an issue for `docker.io` so I have hopefully resolved that...

danieldean:
Ok... hopefully all issues resolved and all good now with the rebase. Bash is deceptively complicated!

src/zwift.sh
```diff
@@ -297,6 +297,21 @@ if [[ ! $DONT_PULL ]]; then
...
+    readarray -t images < <($CONTAINER_TOOL images --filter "reference=${IMAGE/docker.io\/}" --filter "before=${IMAGE/docker.io\/}:$VERSION" --format '{{.ID}}')
+    if [ "${#images[@]}" != "0" ]; then
+        msgbox info "Cleaning up previous container images"
+        (
+            if $CONTAINER_TOOL image rm "${images[@]}"; then
+                msgbox ok "Previous container images have been deleted"
+            else
+                msgbox error "Failed to clean up previous container images"
+            fi
+        ) &
```
netbrain:
```suggestion
            if $CONTAINER_TOOL image rm "${images[@]}"; then
                msgbox ok "Previous container images have been deleted"
            else
                msgbox error "Failed to clean up previous container images"
            fi
```

I've changed my mind, as this might make shell output garbled.

src/zwift.sh
```diff
@@ -297,6 +297,21 @@ if [[ ! $DONT_PULL ]]; then
     fi
 fi
 
+# Clean previous container images (if any)
+if [[ ! $DONT_CLEAN ]] && [[ ! $DONT_PULL ]]; then
+    readarray -t images < <($CONTAINER_TOOL images --filter "reference=${IMAGE/docker.io\/}" --filter "before=${IMAGE/docker.io\/}:$VERSION" --format '{{.ID}}')
+    if [ "${#images[@]}" != "0" ]; then
```
netbrain:
```suggestion
if [ "${#images[@]}" -gt 0 ] && [ -n "${images[0]}" ]; then
```
Is probably a better check

docs/configuration/options.md
```diff
@@ -17,6 +17,7 @@ These environment variables can be used to alter the execution of the zwift bash
 | `VERSION`                   | `latest`                   | The image version/tag to use                                                                                                           |
 | `DONT_CHECK`                |                            | If set, don't check for updated `zwift.sh`                                                                                             |
 | `DONT_PULL`                 |                            | If set, don't pull for new image version                                                                                               |
+| `DONT_CLEAN`                |                            | If set along with `DONT_PULL`, don't clean up previous image                                                                           |
```
netbrain:
```suggestion
| `DONT_CLEAN`                |                            |    If set, don't clean up previous images after pulling                                                                   |
```

src/zwift.sh
```diff
@@ -297,6 +297,21 @@ if [[ ! $DONT_PULL ]]; then
     fi
 fi
 
+# Clean previous container images (if any)
+if [[ ! $DONT_CLEAN ]] && [[ ! $DONT_PULL ]]; then
+    readarray -t images < <($CONTAINER_TOOL images --filter "reference=${IMAGE/docker.io\/}" --filter "before=${IMAGE/docker.io\/}:$VERSION" --format '{{.ID}}')
```
netbrain:
```suggestion
       FILTER_IMAGE="${IMAGE#docker.io/}"
      readarray -t images < <($CONTAINER_TOOL images --filter "reference=${FILTER_IMAGE}" --filter "before=${FILTER_IMAGE}:$VERSION" --format '{{.ID}}')
```
I think this would be better.

 ${var#pattern} only strips from the start of the string, whereas ${var/pattern} replaces the first occurrence anywhere.

src/zwift.sh
```diff
@@ -297,6 +297,21 @@ if [[ ! $DONT_PULL ]]; then
...
+    if [ "${#images[@]}" != "0" ]; then
+        msgbox info "Cleaning up previous container images"
+        (
+            if $CONTAINER_TOOL image rm "${images[@]}"; then
+                msgbox ok "Previous container images have been deleted"
+            else
+                msgbox error "Failed to clean up previous container images"
```
glennvl: Maybe make this a warning instead of error? It's not something that prevents zwift from starting as expected.

src/zwift.sh
danieldean: You'll just need to leave a little more time to get into those races! After the first run with a clean up it should stop being an issue as it'll be at most one image to delete.

danieldean:
Ok, further suggested changes applied. Hope all is good now? I know some more Bash that I did to start with...

docs/configuration/options.md
glennvl:
```suggestion
| `DONT_PULL`                 |                            | If set, don't pull for new image version (implies `DONT_CLEAN`)                                                                        |
| `DONT_CLEAN`                |                            | If set, don't clean up previous image after pulling                                                                                    |
```

Maybe a little improvement to the documentation?

src/zwift.sh
```diff
@@ -297,6 +297,19 @@ if [[ ! $DONT_PULL ]]; then
     fi
 fi
 
+# Clean previous container images (if any)
+if [[ ! $DONT_CLEAN ]] && [[ ! $DONT_PULL ]]; then
+    readarray -t images < <($CONTAINER_TOOL images --filter "reference=${IMAGE#docker.io/}" --filter "before=${IMAGE#docker.io/}:$VERSION" --format '{{.ID}}')
+    if [ "${#images[@]}" -gt 0 ] && [ -n "${images[0]}" ]; then
```
glennvl, line 303:
```suggestion
    if [ "${#images[@]}" -gt 0 ]; then
```

I don't think the `[ -n "${images[0]}" ]` adds anything? I tried with both docker and podman with no images to be deleted. In both cases only the first half of the condition was enough, I did not see an empty first element.

netbrain:
Merging as is, it's fine.

src/zwift.sh
danieldean: In my testing checking for an array with a non-zero length was sufficient as the previous command to get a list of images never returns an empty line (this is what I originally did).

The secondary check may be redundant but given we are deleting images it might be just protect against unforeseen eventualities. Especially as we're not in control of any future changes to the Docker/Podman CLI.

PR #277, opened 2026-02-06
Make linters happy

glennvl:
## Summary

cleaning

Some linters (that are mentioned in the contributors guide) are not yet running in github actions and still produce violations. This PR adds those linters to github actions and fixes violations.

## Implementation

* Add github workflow to lint github actions
* Add github workflow to lint yaml files
* Extend bash script github action to also run shfmt
* Extend nix flake action to also run nixfmt and build a test flake
* Fix violations of the aforementioned linters (inlcuding auto format all sh and yaml files)
* Update devcontainer so all tools are not only available in extensions, but also on the commandline
* Update documentation on how to invoke tools from the commandline

### DIFF
--- a/.github/workflows/lint_bash_scripts.yaml
+++ b/.github/workflows/lint_bash_scripts.yaml
@@ -2,13 +2,9 @@
 name: Validate bash scripts
 on: # yamllint disable-line rule:truthy
   pull_request:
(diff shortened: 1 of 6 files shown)


netbrain:
:+1:

PR #283, opened 2026-02-09
Strict shellcheck (best practices)

glennvl:
## Disclaimer

I considered starting a discussion first, but since I already implemented it in my fork I figured I might as well just open a PR and discuss the proposed changes here if needed.

Me tightening the screws after setting up basic linting. :see_no_evil: 

## Summary

Enable (almost all) optional shellcheck rules:

- **avoid-negated-conditions**
- **check-set-e-suppressed**
- **quote-safe-variables**
- **require-double-brackets**
- **require-variable-braces**

Disabled rules:

- [check-extra-masked-returns](https://github.com/koalaman/shellcheck/wiki/SC2312)
- [check-unassigned-uppercase](https://github.com/koalaman/shellcheck/wiki/SC2154)

### Rules with minor impact on semantics

- **avoid-nullary-conditions**: Don't use `!` to check if a variable is set.

One situation where we encounter violations to this ruls is checks such as `[[ ! $DONT_CHECK ]]`. The rule says either use `-n` or `-eq`. In this PR we went with `[[ $DONT_CHECK -ne "1" ]]`. The behaviour will be a little different to the old version. In the old version setting `DONT_CHECK=0` would still consider DONT_CHECK to be true, in the new version that is not the case.

| `DONT_CHECK` | `[[ ! ${DONT_CHECK} ]]` | `[[ ${DONT_CHECK} -ne "1" ]]` |
|--------------|-------------------------|-------------------------------|
| unset        | true                    | true                          |
| 0            | false                   | true                          |
| 1            | false                   | false                         |
| other value  | false                   | true                          |

The other situation where we run against this rule is checking if the zwift directory is empty with `[[ ! "$(ls -A .)" ]]`. Replacing it with `[[ -z "$(ls -A .)" ]]` fixes the warning and has no impact on semantics.

### Rules with impact on syntax, but no impact on semantics

Rules that required quite a bit of changes:

- **require-double-brackets**: use `[[ ]]` everywhere (safer by default, no further impact on semantics)
- **require-variable-braces**: replace all `$VARIABLE` with `${VARIABLE}`

Rules that required few changes:

- **quote-safe-variables**: only a few violations found, quick and easy fix

### Rules with no impact

No violations were detected for the following rules:

- **avoid-negated-conditions**
- **check-set-e-suppressed**

## Disabled rules

The [google styleguide for shell](https://google.github.io/styleguide/shellguide.html) has some guidelines that could be helpful.

### check-extra-masked-returns

Disallows using commands inside `$()`, conditions, pipelines, ... that are directly passed to another command, since it hides the exit status. A rule that we should probably enable at some point, but we need to implement better error handling first. We probably also need to evaluate using `set -e` and `set -o pipefail`. A job for a different PR.

### check-unassigned-uppercase

Checks that variables are assigned a value before they are used. For starters we are using uppercase variables everywhere, violating the convention of using lowercase for local variables and uppercase for exported variables. Probably something we should fix, but in a different PR so this one does not become too large. Then we can also make sure everything is initialized with default values and enable this rule.

## Bash code in github actions

Actionlint invokes shellcheck on bash code in github actions. Sadly it [forces its own configuration options](https://github.com/rhysd/actionlint/blob/main/rule_shellcheck.go#L192).

```go
args := []string{"--norc", "-f", "json", "-x", "--shell", sh, "-e", "SC1091,SC2194,SC2050,SC2154,SC2157,SC2043", "-"}
```

The `--norc` makes shellcheck ignore the `.shellcheckrc` file, so it's not possible to pass a custom configuration. In addition the `-e` disables a bunch of rules: SC1091, SC2194, SC2050, SC2154, SC2157 and SC2043.

We have two options here:

1. Accept that the bash code in github actions is not checked as thoroughly.
2. Move the bash code in github actions to scripts that are invoked by the workflows instead.

Again, not something for this PR.

### DIFF
--- a/src/zwift.sh
+++ b/src/zwift.sh
@@ -219,6 +219,11 @@ else
     msgbox warning "No Zwift credentials found..."
 fi
 
+# Pass environment variable to container if gamemode should be disabled
+if [[ ${ZWIFT_NO_GAMEMODE} -eq "1" ]]; then
+    ENVIRONMENT_VARIABLES+=(ZWIFT_NO_GAMEMODE="1")
+fi
+
 if [[ ${CONTAINER_TOOL} == "podman" ]]; then
     # Podman has to use container id 1000
     # Local user is mapped to the container id


netbrain:
I believe we can accept some code to not be checked as thoroughly. And if proven problematic in the future, we can extract the problematic code into their own scripts. Let's try to follow the kiss (keep it simple stupid) principle.

PR #286, opened 2026-02-10
Enable shellcheck check-unassigned-uppercase rule

glennvl:
## Disclaimer

A lot of code was changed in this PR. While these changes **should not** affect functionality, thorough testing before merging is advised.

## Summary

This PR enables the [check-unassigned-uppercase](https://github.com/koalaman/shellcheck/wiki/SC2154) shellcheck rule.

The following changes were made to fix violations against this rule:

* Initialize environment variables with a default value
* Move constants to the top of scripts and declare them as readonly
* Mark local variables in functions as such and convert them to lower case
* Convert all other local variables in scripts to lower case

## Implementation

Some additional changes were made to make the above list easier and to take into account additional best practices for bash scripting:

* Always use quotes, except for literal numbers
* Replace `$@`, `$1`, `$2`, `$3` with `${@}`, `${1}`, `${2}`, `${3}` respectively
* Use a single array for container arguments instead of many variables in `zwift.sh` (and also `build-image.sh`)
* Reorder code in `zwift.sh` to be more logical
* Enable `set -x` in all scripts based on the `DEBUG` environment variable (which is now also propagated to the container)

## Bugfixes

While making these changes I encountered two defects in `entrypoint.sh`, it is unlikely that they impact anyone:

1. `GID` was used instead of `UID` to create the run directory if it does not exist. Fixed in 76bb33b2a7ae4d677bffe34dda2f1dccd82d6fd7.
2. The check to see if the `UID` and `GID` are valid numbers always returned true. Fixed in 8877a982e9d6ea8b774d4e8bbc21bf818d0a23dd.
3. The invalid uid/gid message was also shown if the ids are valid and the same as the user ids. Fixed in 8ebe6a5ce2d6f253896cebfbd3c52e50ba39f61f.

Other fixed issues:

1. Setting `WINE_EXPERIMENTAL_WAYLAND=1` causes zwift to fail to start with an opengl error. This can be solved by installing `libegl1` in the Dockerfile. Fixed in 0a51e2b4d91a1728e91f3e8d8a1b5d76fe499556.

## Other potential issues (not fixed yet)

I also discovered a few more potential issues:

1. In `update_zwift.sh`, if zwift is not installed yet (`ZWIFT_HOME` directory empty) **and** the update flag is set, the `wait_for_zwift_game_update` function is executed twice.
2. Also in `update_zwift.sh`, if zwift is not installed yet (`ZWIFT_HOME` directory empty), `ZwiftLauncher.exe` is launched after `ZwiftSetup.exe` **and** then launched again in the background by `wait_for_zwift_game_update`.
3. In `zwift.sh` the positional arguments `${@}` are passed to the container tool before `IMAGE:VERSION`. This way they function as additional arguments to podman or docker. Is this intentional or are they meant to be passed as command to the container (`IMAGE:VERSION` first and then `${@}`)?
6. In `entrypoint.sh`, if `CONTAINER_TOOL` is not set, it automatically tries to detect whether the container is hosted by docker or podman. The docker check works (a `./dockerenv` file is indeed only present in case of docker). But the podman check does not. It tries to look for `podman` inside a `/run/.containerenv` file. That file is indeed present in case of podman (and not present in case of docker), but it is empty. So if `CONTAINER_TOOL` is not set and the tool is podman, it will fail with the error `unknown container`.
7. While playing with `CONTAINER_EXTRA_ARGS` I found values with spaces break the final command. For example `CONTAINER_EXTRA_ARGS="--env MY_VAR=value --cpus=1.5"` is split into `("--env", "MY_VAR=value", "--cpus=1.5")`, which works exactly as it should. But let's say I want to give `MY_VAR` a value that contains spaces. `CONTAINER_EXTRA_ARGS="--env MY_VAR='this has spaces'"` is split into `("--env", "MY_VAR='this", "has", "spaces'")`, this breaks the final command and prevents the container from starting. The root cause is the `read -ra` call to split the command, this was the case in the original version of the script and is still used in this version. The `VGA_DEVICE_FLAG` parameter has the same issue.

   ```bash
   readonly CONTAINER_EXTRA_ARGS="--env a='b c'"
   read -ra extra_args <<< "${CONTAINER_EXTRA_ARGS}"
   declare -p extra_args
   # output: declare -a extra_args=([0]="--env" [1]="a='b" [2]="c'")
   ```

   The splitting is required to make things work, so we can't just leave it out. But the splitting done by `read -ra` is dumb, it splits on spaces and newlines disregarding quotes to keep stuff together. `readarray` and `mapfile` do an even worse job, so they are not a solution.

Fixing these issues:

* The first two require a complete refactoring of `update_zwift.sh`, so they should probably get their own PR (after this one is merged to avoid conflicts).
* The third one can easily be fixed in this PR if desired.
* As for the fourth one, I think we should not try to automatically detect which container tool is being used inside the container. We always pass the `CONTAINER_TOOL` environment variable, so I think we should just exit with an error if it is not set. (If the automatic detection worked, it would potentially hide a bug of a missing `CONTAINER_TOOL` variable.) This can also be easily fixed in this PR if desired.
* I'm not sure how to fix the fifth one. Maybe telling the user to declare `CONTAINER_EXTRA_ARGS` and `VGA_DEVICE_FLAG` as an array instead of a string is the way to go.

### Tested

* [x] Test building image with podman
* [x] Test building image with docker
* [x] Test running zwift with podman
* [x] Test running zwift with docker
* [x] Test various configuration options
  * [x] DEBUG="1"
  * [x] USER="bob"
  * [x] CONTAINER_TOOL="podman"
  * [x] CONTAINER_TOOL="docker"
  * [x] IMAGE="localhost/zwift"
  * [x] VERSION="latest"
  * [x] DONT_CHECK="1"
  * [x] DONT_PULL="1"
  * [x] ZWIFT_FG="1"
  * [x] INTERACTIVE="1"
  * [x] DRYRUN="1" 
  * [x] ZWIFT_USERNAME="..."
  * [x] ZWIFT_PASSWORD="..."
  * [x] ZWIFT_WORKOUT_DIR="$(xdg-user-dir DOCUMENTS)/Zwift/Workouts"
  * [x] ZWIFT_ACTIVITY_DIR="$(xdg-user-dir DOCUMENTS)/Zwift/Activities"
  * [x] ZWIFT_LOG_DIR="$(xdg-user-dir DOCUMENTS)/Zwift/Logs"
  * [x] ZWIFT_SCREENSHOTS_DIR="$(xdg-user-dir PICTURES)/Zwift"
  * [x] NETWORKING="host"
  * [x] ZWIFT_OVERRIDE_GRAPHICS="1"
  * [x] ZWIFT_OVERRIDE_RESOLUTION="1920x1080"
  * [x] WINE_EXPERIMENTAL_WAYLAND="1"
  * [x] CONTAINER_EXTRA_ARGS="--env MY_ENV_VAR=value --cpus=1.5"
  * [x] VGA_DEVICE_FLAG="--device=nvidia.com/gpu=all"
  * [x] PRIVILEGED_CONTAINER="1"
  * [x] ZWIFT_NO_GAMEMODE="1"

### Not Tested

* ZWIFT_UID
* ZWIFT_GID

## Next up

The next PR in this series will look at enabling the [check-extra-masked-returns](https://github.com/koalaman/shellcheck/wiki/SC2312) shellcheck rule:

* Add missing error handling and improve existing error handling
* Decide if enabling `set -e` in all scripts is a good idea
* Decide if enabling `set -o pipefail` in all scripts is a good idea

### DIFF
--- a/src/build-image.sh
+++ b/src/build-image.sh
@@ -39,6 +39,8 @@ container_args=(
     -e DISPLAY="${DISPLAY}"
     -e XAUTHORITY="${XAUTHORITY}"
     -e CONTAINER_TOOL="${CONTAINER_TOOL}"
+    -e ZWIFT_UID="${ZWIFT_UID}"
+    -e ZWIFT_GID="${ZWIFT_GID}"
 
     -v /tmp/.X11-unix:/tmp/.X11-unix
     -v "/run/user/${UID}:/run/user/${ZWIFT_UID}"
--- a/src/entrypoint.sh
+++ b/src/entrypoint.sh
@@ -4,8 +4,8 @@ set -e
 readonly DEBUG="${DEBUG:-0}"
 if [[ ${DEBUG} -eq 1 ]]; then set -x; fi
 
-readonly ZWIFT_UID=${ZWIFT_UID:-}
-readonly ZWIFT_GID=${ZWIFT_GID:-}
+readonly ZWIFT_UID=${ZWIFT_UID:-$(id -u user)}
+readonly ZWIFT_GID=${ZWIFT_GID:-$(id -g user)}
 readonly WINE_EXPERIMENTAL_WAYLAND=${WINE_EXPERIMENTAL_WAYLAND:-0}
 
 readonly ZWIFT_HOME="/home/user/.wine/drive_c/Program Files (x86)/Zwift"
@@ -50,17 +50,17 @@ if [[ ${CONTAINER_TOOL} == "docker" ]]; then
 
     # Test that it exists and is a number, and only if different to existing.
     switch_user_id=0
-    if [[ -n ${ZWIFT_UID} ]] && [[ ${ZWIFT_UID} -eq ${ZWIFT_UID} ]] && [[ ${user_uid} -ne ${ZWIFT_UID} ]]; then
+    if [[ ! ${ZWIFT_UID} =~ ^[0-9]+$ ]]; then
+        echo "ZWIFT_UID is not a number: '${ZWIFT_UID}'" >&2
+    elif [[ ${user_uid} -ne ${ZWIFT_UID} ]]; then
         user_uid="${ZWIFT_UID}"
         switch_user_id=1
-    else
-        echo "ZWIFT_UID is not set or not a number: '${ZWIFT_UID}'"
     fi
-    if [[ -n ${ZWIFT_GID} ]] && [[ ${ZWIFT_GID} -eq ${ZWIFT_GID} ]] && [[ ${user_gid} -ne ${ZWIFT_GID} ]]; then
+    if [[ ! ${ZWIFT_GID} =~ ^[0-9]+$ ]]; then
+        echo "ZWIFT_GID is not a number: '${ZWIFT_GID}'" >&2
+    elif [[ ${user_gid} -ne ${ZWIFT_GID} ]]; then
         user_gid="${ZWIFT_GID}"
         switch_user_id=1
-    else
-        echo "ZWIFT_GID is not set or not a number: '${ZWIFT_GID}'"
     fi
 
     # This section is only run if we are switching either UID or GID.


glennvl:
@netbrain I just did a lot more testing and fixed a few final things. I think this PR is ready for review now. :sweat_smile:

netbrain:
I'll take a look in due time! 😅

glennvl:
I've been experimenting a bit with the `CONTAINER_EXTRA_ARGS` (and `VGA_DEVICE_FLAG`) issue, and came up with the following possible solution:

* Recommend the user to declare those two variables as an array, it's the only way that works in all cases
* Either keep handling the string case as we do now or exit with an error message telling the user to convert to array

```bash
#!/usr/bin/env bash

if [[ -f "./config" ]]; then
    # shellcheck source=/dev/null
    source "./config"
fi

readonly CONTAINER_EXTRA_ARGS="${CONTAINER_EXTRA_ARGS:-}" # initialize as empty string if not set

is_array() {
    local variable_name="${1}"
    local array_regex="^declare -a"
    [[ "$(declare -p "${variable_name}")" =~ ${array_regex} ]]
}

declare -a container_args=()

if is_array "CONTAINER_EXTRA_ARGS"; then
    # note: adds an empty value if the array is empty, but that does not affect the final command
    container_args+=("${CONTAINER_EXTRA_ARGS[@]}")
elif [[ -n ${CONTAINER_EXTRA_ARGS} ]]; then
    # the same as we do now
    read -ra extra_args <<< "${CONTAINER_EXTRA_ARGS}"
    container_args+=("${extra_args[@]}")
    # or exit with error
    # msgbox error "Convert CONTAINER_EXTRA_ARGS to array ..."
    # exit 1
fi

declare -p container_args
```

* no extra args

   ```console
   foo@bar:~$ ./test.sh
   declare -a container_args=()
   ```

* extra args string (can be passed on commandline or through config file)

   ```console
   foo@bar:~$ CONTAINER_EXTRA_ARGS="--some_arg --env A=B --cpus=1.5" ./test.sh
   declare -a container_args=([0]="--some_arg" [1]="--env" [2]="A=B" [3]="--cpus=1.5")
   foo@bar:~$ echo "CONTAINER_EXTRA_ARGS=\"--some_arg --env A=B --cpus=1.5\"" > ./config
   foo@bar:~$ ./test.sh
   declare -a container_args=([0]="--some_arg" [1]="--env" [2]="A=B" [3]="--cpus=1.5")
   ```

* extra args array (cannot declare on the commandline, but works in sourced config file)

   ```console
   foo@bar:~$ echo "CONTAINER_EXTRA_ARGS=(--some_arg --env A=B --cpus=1.5)" > ./config
   foo@bar:~$ ./test.sh
   declare -a container_args=([0]="--some_arg" [1]="--env" [2]="A=B" [3]="--cpus=1.5")
   ```

* extra args string with value that contains spaces (not split correctly, this is the bug that we're trying to fix)

   ```console
   foo@bar:~$ CONTAINER_EXTRA_ARGS='--env a="b c" --env d=e --cpu=1.5' ./test.sh
   declare -a container_args=([0]="--env" [1]="a=\"b" [2]="c\"" [3]="--env" [4]="d=e" [5]="--cpu=1.5")
   foo@bar:~$ echo "CONTAINER_EXTRA_ARGS='--env a=\"b c\" --env d=e --cpu=1.5'" > ./config
   foo@bar:~$ ./test.sh
   declare -a container_args=([0]="--env" [1]="a=\"b" [2]="c\"" [3]="--env" [4]="d=e" [5]="--cpu=1.5")
   ```

* extra args array with value that containes spaces (works just fine)

   ```console
   foo@bar:~$ echo "CONTAINER_EXTRA_ARGS=(--env a=\"b c\" --env d=e --cpu=1.5)" > ./config
   foo@bar:~$ ./test.sh
   declare -a container_args=([0]="--env" [1]="a=b c" [2]="--env" [3]="d=e" [4]="--cpu=1.5")
   foo@bar:~$ echo "CONTAINER_EXTRA_ARGS=(--env a='b c' --env d=e --cpu=1.5)" > ./config
   foo@bar:~$ ./test.sh
   declare -a container_args=([0]="--env" [1]="a=b c" [2]="--env" [3]="d=e" [4]="--cpu=1.5")
   ```

glennvl:
@netbrain

Damn, you already went for it, I didn't expect this to get merged without me having to change some stuff haha. :sweat_smile: 

What do we do with the list of issues I discovered that are not fixed yet? Should I create a single issue for all of them where we can discuss them? Or create separate issues for each of them?

netbrain:
Follow up in new issue

PR #291, opened 2026-02-13
Entrypoint script container tool detection

### LINKED ISSUE #289: Automatic ontainer tool detection in entrypoint does not work for podman
## Describe the bug

First mentioned in other potential issues in #286.

> 4. In `entrypoint.sh`, if `CONTAINER_TOOL` is not set, it automatically tries to detect whether the container is hosted by docker or podman. The docker check works (a `./dockerenv` file is indeed only present in case of docker). But the podman check does not. It tries to look for `podman` inside a `/run/.containerenv` file. That file is indeed present in case of podman (and not present in case of docker), but it is empty. So if `CONTAINER_TOOL` is not set and the tool is podman, it will fail with the error `unknown container`.

## Expected behaviour

Either detect the container tool correctly or don't attempt to detect it at all and show an error if it is not set.

## Possible solution

I think we should not try to automatically detect which container tool is being used inside the container. We always pass the `CONTAINER_TOOL` environment variable, so I think we should just exit with an error if it is not set. If the automatic detection worked, it would potentially hide a bug of a missing `CONTAINER_TOOL` variable.

glennvl:
Fixes #289.

* Remove automatic container tool detection, since it was incorrect/unreliable.
* Fix `CONTAINER_TOOL` environment variable name in github workflows.

### DIFF
--- a/src/entrypoint.sh
+++ b/src/entrypoint.sh
@@ -7,25 +7,10 @@ if [[ ${DEBUG} -eq 1 ]]; then set -x; fi
 readonly ZWIFT_UID="${ZWIFT_UID:-$(id -u user)}"
 readonly ZWIFT_GID="${ZWIFT_GID:-$(id -g user)}"
 readonly WINE_EXPERIMENTAL_WAYLAND="${WINE_EXPERIMENTAL_WAYLAND:-0}"
+readonly CONTAINER_TOOL="${CONTAINER_TOOL:?}"
 
 readonly ZWIFT_HOME="/home/user/.wine/drive_c/Program Files (x86)/Zwift"
 readonly ZWIFT_DOCS="/home/user/.wine/drive_c/users/user/Documents/Zwift"
-
-# Check whether we are running in Docker/ Podman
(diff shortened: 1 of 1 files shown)


netbrain:
This will enforce all users to have to specify a container tool env? or can we default to say docker? and podman users will have to override?

glennvl:
> This will enforce all users to have to specify a container tool env? or can we default to say docker? and podman users will have to override?

This only changes things inside the container. The zwift.sh script always sets the `CONTAINER_TOOL` environment variable, so the detection inside the container is not being used unless someone manually starts the container, bypassing the zwift.sh script, without passing the necessary parameters.

The zwift.sh script still does the same as before. If a `CONTAINER_TOOL` is set by the user, try to use that. If no `CONTAINER_TOOL` is set, check if podman is installed, if so, use it, if not, try to use docker. If the container tool is not found, show an error message with info how to install podman/docker.

netbrain:
Ah right, understod.

netbrain:
LGTM

PR #295, opened 2026-02-17
Rebuild on dockerfile change

### LINKED ISSUE #294: Zwift updater workflow does not handle dockerfile changes
## Describe the bug

Creating a separate issue for https://github.com/netbrain/zwift/issues/287#issuecomment-3895956085.

When a change is made to the dockerfile, the container image is not automatically updated.

* The zwift updater workflow does not handle the case of an updated dockerfile. It only updates the scripts in the container image.
* The zwift rebuild from scratch workflow only runs once every three months or when manually triggered.

> Yeah, that's correct, some things the zwift updater can't handle, i.e Dockerfile changes. in this case it would make sense to have build from scratch trigger from push based on a file filter absolutely. However in order to keep complexity down, i've opted to only do this manually for now. But by all means, i think including the Dockerfile on push event for this is probably a good thing.
> 
> And you probably want to stick to using the existing workflows, no new workflow required. just add on push and file filter.

glennvl:
Fixes #294 

Trigger build from scratch workflow on dockerfile changes.


glennvl:
@netbrain Is the `squash.Dockerfile` actually used anywhere or needed for something? Should we keep it around or can we delete it?

netbrain:
No I don't think this is used anymore. Used to have a squash workflow, but ended up just building from scratch in the end instead. 

Feel free to delete.

netbrain (approved):

PR #297, opened 2026-02-20
Merge linting workflows

glennvl:
## Summary

This PR merges the 6 separate linter workflow files into a single workflow file.

* They all had the same trigger anyway
* Reduce the amount of spam in the actions tab (one entry instead of six for each trigger)
* Make it possible to create a linter workflow badge in the README

[Lint everything](https://github.com/netbrain/zwift/actions/workflows/lint_everything.yaml)

## Discussion

We are always running all linters, regardless of the files that have changed. I think this is fine, but if we wanted to we could run linters depending on which files changed using https://github.com/dorny/paths-filter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # Zwift
 
+[![Lint everything][lint-everything-src]][lint-everything-href]
 [![Zwift updater][zwift-updater-src]][zwift-updater-href]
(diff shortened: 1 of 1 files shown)


netbrain:
lgtm!

PR #298, opened 2026-02-20
Strict lint bash in workflows

glennvl:
## Summary

Apply the same rules to the bash code in github actions as we do for shell scripts.

## Implementation

Because strict bash code linting for actions can't be automated (actionlint runs shellcheck with less strict rules), I followed the following steps:

- Copy bash code from github workflow to temporary script
- Run shellcheck on the script
- Fix any violations in the script
- Copy code from the script back to the github workflow

I also removed the `set -eo pipefail` because that already is [the default](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defaultsrunshell).

### DIFF
--- a/.github/workflows/zwift_build_from_scratch.yaml
+++ b/.github/workflows/zwift_build_from_scratch.yaml
@@ -48,7 +48,7 @@ jobs:
             -e DISPLAY="${DISPLAY}" \
             -e CONTAINER_TOOL="docker" \
             --device="/dev/dri:/dev/dri" \
-            -v "/tmp:/tmp" \
+            -v /tmp/.X11-unix:/tmp/.X11-unix \
             netbrain/zwift:build
 
         docker commit \
--- a/.github/workflows/zwift_build_matrix.yaml
+++ b/.github/workflows/zwift_build_matrix.yaml
@@ -62,5 +62,5 @@ jobs:
             -e DISPLAY="${DISPLAY}" \
             -e CONTAINER_TOOL="docker" \
             --device="/dev/dri:/dev/dri" \
-            -v "/tmp:/tmp" \
+            -v -v /tmp/.X11-unix:/tmp/.X11-unix \
             netbrain/zwift:build
--- a/.github/workflows/zwift_updater.yaml
+++ b/.github/workflows/zwift_updater.yaml
@@ -47,7 +47,7 @@ jobs:
             -e DISPLAY="${DISPLAY}" \
             -e CONTAINER_TOOL="docker" \
             --device="/dev/dri:/dev/dri" \
-            -v "/tmp:/tmp" \
+            -v /tmp/.X11-unix:/tmp/.X11-unix \
             netbrain/zwift:latest update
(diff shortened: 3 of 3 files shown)


.github/workflows/zwift_build_from_scratch.yaml
```diff
@@ -29,33 +29,36 @@ jobs:
...
-                -e CONTAINER_TOOL=docker \
-                --device /dev/dri \
-                -v /tmp:/tmp \
-                netbrain/zwift:build
+        docker run \
+            --name zwift \
+            -e DISPLAY="${DISPLAY}" \
+            -e CONTAINER_TOOL="docker" \
+            --device="/dev/dri:/dev/dri" \
+            -v "/tmp:/tmp" \
```
glennvl: Why are we mounting `/tmp`?

.github/workflows/zwift_build_matrix.yaml
```diff
@@ -42,26 +42,25 @@ jobs:
...
-                -e CONTAINER_TOOL=docker \
-                --device /dev/dri \
-                -v /tmp:/tmp \
-                netbrain/zwift:build
+        docker run \
+            --name zwift \
+            -e DISPLAY="${DISPLAY}" \
+            -e CONTAINER_TOOL="docker" \
+            --device="/dev/dri:/dev/dri" \
+            -v "/tmp:/tmp" \
```
glennvl: Why are we mounting `/tmp`?

.github/workflows/zwift_updater.yaml
```diff
@@ -22,39 +22,42 @@ jobs:
...
-          "netbrain/zwift:$VERSION"
-
-        docker tag "netbrain/zwift:$VERSION" netbrain/zwift:latest
-        docker push "netbrain/zwift:$VERSION"
+        docker run \
+            --name zwift \
+            -e DISPLAY="${DISPLAY}" \
+            -e CONTAINER_TOOL="docker" \
+            --device="/dev/dri:/dev/dri" \
+            -v "/tmp:/tmp" \
```
glennvl: Why are we mounting `/tmp`?

netbrain: It might require  /tmp/.X11-unix/, or some other tmp related file. not sure.

.github/workflows/zwift_build_matrix.yaml
netbrain: It might require  /tmp/.X11-unix/, or some other tmp related file. not sure.

.github/workflows/zwift_build_from_scratch.yaml
netbrain: It might require  /tmp/.X11-unix/, or some other tmp related file. not sure.

glennvl: That would make sense, so should we just do `-v /tmp/.X11-unix:/tmp/.X11-unix` instead?

PR #300, opened 2026-02-22
Warn if updates are disabled

glennvl:
Show a warning if DONT_CHECK or DONT_PULL are set. For context, see https://github.com/netbrain/zwift/discussions/299#discussioncomment-15875468.


src/zwift.sh
```diff
@@ -172,6 +174,8 @@ if [[ ${DONT_PULL} -ne 1 ]]; then
     else
         msgbox error "Failed to update container image"
     fi
+else
+    msgbox warning "Not checking for new container image"
```
netbrain: Maybe add," you have to check for updates manually". And/Or "should only be used if latest image is not working for you”

glennvl: I'll think about better messages tomorrow.

glennvl: Updated messages in fb62bde

PR #303, opened 2026-02-26
Add configuration parameter for extra environment variables

glennvl:
## Summary

Make it a bit more convenient for the end user to pass extra environment variables to the container by adding a `CONTAINER_EXTRA_ENVVARS` option.

Example usage in `~/.config/zwift`:

```bash
CONTAINER_EXTRA_ENVVARS=(
    XCURSOR_SIZE=48          # increase cursor size
    MESSAGE="hello, world!"  # some other environment variable called MESSAGE
)
```

This is merely a convenience option for the end user, as it is equivalent to adding:

```bash
CONTAINER_EXTRA_ARGS=(
    -e XCURSOR_SIZE=48
    -e MESSAGE="hello, world"
)
```

## Discussion

I implemented this a while ago, so I figured I might as well create a pull request for it. Alternatively we could just update the documentation, add how to pass environment variables using CONTAINER_EXTRA_ARGS.

In my opinion adding CONTAINER_EXTRA_ENVVARS has value since it allows the user to beter express intent compared to adding them to CONTAINER_EXTRA_ARGS.

### DIFF
--- a/docs/configuration/options.md
+++ b/docs/configuration/options.md
@@ -22,6 +22,7 @@ These environment variables can be used to alter the execution of the zwift bash
 | `INTERACTIVE`               | `0`                        | If set to `1`, force `-it` and use `--entrypoint bash` for debugging                                                                   |
 | `CONTAINER_TOOL`            |                            | Defaults to podman if installed, else docker                                                                                           |
 | `CONTAINER_EXTRA_ARGS`      |                            | Extra args passed to docker/podman (`--cpus=1.5`)                                                                                      |
+| `CONTAINER_EXTRA_ENVVARS`   |                            | Extra environment variables passed to the container (`KEY="value"`)                                                                    |
 | `ZWIFT_USERNAME`            |                            | If set, try to login to zwift automatically                                                                                            |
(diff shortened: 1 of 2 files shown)


netbrain:
I think this is already covered by `CONTAINER_EXTRA_ARGS`:

```bash
CONTAINER_EXTRA_ARGS=(-e XCURSOR_SIZE=48 -e MESSAGE="hello, world!")
```

I'd rather keep things simple and not add another config option for something that's already possible. Maybe we could just add a documentation example showing how to pass env vars with `CONTAINER_EXTRA_ARGS` instead?

glennvl:
> I think this is already covered by `CONTAINER_EXTRA_ARGS`:
> 
> ```shell
> CONTAINER_EXTRA_ARGS=(-e XCURSOR_SIZE=48 -e MESSAGE="hello, world!")
> ```
> 
> I'd rather keep things simple and not add another config option for something that's already possible. Maybe we could just add a documentation example showing how to pass env vars with `CONTAINER_EXTRA_ARGS` instead?

Yup, all good to me. I created the pull request since I already implemented it anyway a while ago. So I figured I'd let you decide what to do with it so I can clean up this branch on my fork. :wink: 

You can close this PR if you want. The documentation can be updated in a different one.

PR #302, opened 2026-02-25
Migrate Zwift data to from Documents/Zwift to AppData/Local/Zwift

glennvl:
Fixes #301

~Draft until properly tested and confirmed that it is safe to overwrite existing data in AppData/Local/Zwift.~

PR #304, opened 2026-02-26
Add SCRIPT_VERSION option

glennvl:
## Summary

This PR introduces a new configuration option `SCRIPT_VERSION`. Similar to `VERSION` for the container image, it allows the user to pin the script to a specific version.

* By default `SCRIPT_VERSION` is set to the value `master`, so the latest version of the script is used.
* The user can then set `SCRIPT_VERSION` to a commit hash to use a specific version of the script. As is usual for git, it is enough to use the first 6 digits of the commit hash, by any amount between 6 and the full commit hash is allowed.

## Reasoning

If a script update causes zwift to be unable to launch for a user, there is currently no easy way for the user to be able to zwift until the issue is resolved. With `SCRIPT_VERSION` he can rollback to an earlier version of the script and still be able to zwift while the issue is being resolved.

## Usage

Look at the history of the zwift script and pick a commit id.

* History: <https://github.com/netbrain/zwift/commits/master/src/zwift.sh>
* Set the desired script version either in the `~/.config/zwift/` config file or pass it on the command line 

```bash
SCRIPT_VERSION="44dccc" # pin zwift.sh script to commit 44dccc65f253aba78b67268e119bd31ca63a9efd
```

Or more likely after reporting an issue that zwift is no longer working since a script update, set the `SCRIPT_VERSION` to the commit id recommended by the response in the issue until a new version of the script is released.

## Implementation

* Add a `SCRIPT_VERSION` configuration option
* Add a commandline argument to the install script to specificy the script version
* Replace the script url with `https://raw.githubusercontent.com/netbrain/zwift/${SCRIPT_VERSION}/src/zwift.sh` in both the zwift script and the install script

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -4,12 +4,15 @@ readonly DEBUG="${DEBUG:-0}"
 if [[ ${DEBUG} -eq 1 ]]; then set -x; fi
 
 print_usage() {
-    echo "Usage: ${0} [ -v | --script-version COMMIT_HASH ]"
+    echo "Usage: install [ -v | --script-version COMMIT_HASH ]"
+    echo "               [ -y | --auto-confirm ]"
+    echo "               [ -h | --help ]"
     exit 2
 }
 
 script_version="master"
-if ! options="$(getopt -o "v:h" -l "script-version:,help" -- "${@}")"; then
+auto_confirm=0
+if ! options="$(getopt -n install -o "v:yh" -l "script-version:,auto-confirm,help" -- "${@}")"; then
     print_usage
 fi
 eval set -- "${options}"
@@ -19,6 +22,10 @@ while :; do
             script_version="${2}"
             shift 2
             ;;
+        -y | --auto-confirm)
+            auto_confirm=1
+            shift
+            ;;
         -h | --help)
             print_usage
             ;;
@@ -27,7 +34,8 @@ while :; do
             break
             ;;
         *)
-            exit 1
+            echo "Unexpected option: ${1} - this should not happen." >&2
+            print_usage
             ;;
     esac
 done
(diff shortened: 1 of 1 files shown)


src/zwift.sh
```diff
@@ -105,6 +105,8 @@ load_config_file "${USER_CONFIG_DIR}/${USER}-config"
 # Initialize environment variables
 readonly IMAGE="${IMAGE:-docker.io/netbrain/zwift}"
 readonly VERSION="${VERSION:-latest}"
+readonly LATEST_SCRIPT_VERSION="master"
+readonly SCRIPT_VERSION="${SCRIPT_VERSION:-${LATEST_SCRIPT_VERSION}}"
```
netbrain, line 109:
```suggestion
readonly SCRIPT_VERSION="${SCRIPT_VERSION:-master}"
```

netbrain (commented):
👍

src/zwift.sh
glennvl: The reason for the LATEST_SCRIPT_VERSION variable is so we can compare SCRIPT_VERSION to latest to issue a warning if a custom version is being used. Then we don't need to use the hardcoded value `master` anywhere.

glennvl: Or we could replace LATEST_SCRIPT_VERSION with master everywhere. Whatever you prefer.

netbrain: Ah i didn't see it was actually used elsewhere.

keesjan42:
I just started and did at first NOT update the script. Then I saw that the container was pulled (I'm sure this phrasing is wrong...) and then it said that Zwift is launched, but nothing happened.
I waited a few minutes and then I stopped the terminal window (I always start in a terminal window) and started again and then DID update the script by typing "Y". Then everything started as it should.
Is this normal behaviour?

glennvl:
> I just started and did at first NOT update the script. Then I saw that the container was pulled (I'm sure this phrasing is wrong...) and then it said that Zwift is launched, but nothing happened. I waited a few minutes and then I stopped the terminal window (I always start in a terminal window) and started again and then DID update the script by typing "Y". Then everything started as it should. Is this normal behaviour?

It is normal if you didn't update to #302 yet either. That one was merged yesterday and has changes in both the script and image that would cause zwift to fail to launch with an old script version.

This PR did not change anything to the image, only to the script. So ignoring this update would keep everything as is.

Please create an issue next time instead of commenting in a merged PR. :slightly_smiling_face:

PR #307, opened 2026-02-26
Fix DISPLAY not exported in build workflows

netbrain:
## Summary
- `readonly` does not export variables to the environment, causing `xhost` and `docker run` to not see `DISPLAY`
- The updater workflow was already fixed directly on master (b297551), this PR fixes the same issue in:
  - `zwift_build_matrix.yaml`
  - `zwift_build_from_scratch.yaml`
- Changes `readonly DISPLAY=":99"` to `export DISPLAY=":99"`

## Context
Introduced in 75019e6 ("Strict lint bash in workflows") which mechanically replaced `export` with `readonly`.

### DIFF
--- a/.github/workflows/zwift_build_from_scratch.yaml
+++ b/.github/workflows/zwift_build_from_scratch.yaml
@@ -29,7 +29,7 @@ jobs:
       uses: actions/checkout@v6
     - name: Build zwift
       run: |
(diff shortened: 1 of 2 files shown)


glennvl:
Thanks for fixing the stuff I broke :sweat_smile:

PR #309, opened 2026-02-28
Improve error handling

### LINKED ISSUE #290: Conflicts between install and update in update_zwift script
## Describe the bug

Opening an issue to not forget about this. Something that we should look at fixing at some point, but not urgent.

First mentioned in #286.

> 1. In `update_zwift.sh`, if zwift is not installed yet (`ZWIFT_HOME` directory empty) and the update flag is set, the `wait_for_zwift_game_update` function is executed twice.
>
> 2. Also in `update_zwift.sh`, if zwift is not installed yet (`ZWIFT_HOME` directory empty), `ZwiftLauncher.exe` is launched after `ZwiftSetup.exe` and then launched again in the background by `wait_for_zwift_game_update`.

## Expected behavior

* No conflicts between install and update.
* Launcher only executed once?

## Possible solution

Refactor update_zwift script.

glennvl:
## Summary

Here's the final big PR. This one adds proper error handling to all scripts.

- Enable final shellcheck rule [check-extra-masked-return](https://www.shellcheck.net/wiki/SC2312)
- Refactor update script, fixes #290
- Auth script: Enable `set -euo pipefail`, exit if anything goes wrong and handle failure in the launch script
- Other scripts: Enable `set -uo pipefail`, check the exit code of all commands and act accordingly

## Implementation

Enabled bash flags:

- `set -e` causes the script to exit if a command returns non-zero
- `set -u` causes the script to exit if a variable that has not been set is accessed
- `set -o pipefail` propagates errors in pipelines so they are not hidden

For context why not to use `set -e` in other scripts, see: <https://mywiki.wooledge.org/BashFAQ/105>.

What it boils down to is:

- `set -e` is automatically disabled inside functions
- expressions return the value of the operator `i=1; ((i++))` would return non-zero, causing the script to exit

### DIFF
--- a/src/build-image.sh
+++ b/src/build-image.sh
@@ -5,6 +5,7 @@ readonly DEBUG="${DEBUG:-0}"
 if [[ ${DEBUG} -eq 1 ]]; then set -x; fi
 
 if [[ -t 1 ]]; then
+    readonly COLORED_OUTPUT_SUPPORTED="1"
     readonly COLOR_WHITE="\033[0;37m"
     readonly COLOR_RED="\033[0;31m"
     readonly COLOR_GREEN="\033[0;32m"
@@ -14,6 +15,7 @@ if [[ -t 1 ]]; then
     readonly STYLE_UNDERLINE="\033[4m"
     readonly RESET_STYLE="\033[0m"
 else
+    readonly COLORED_OUTPUT_SUPPORTED="0"
     readonly COLOR_WHITE=""
     readonly COLOR_RED=""
     readonly COLOR_GREEN=""
@@ -104,7 +106,7 @@ container_args=(
     --hostname "${HOSTNAME}"
 
     -e DEBUG="${DEBUG}"
-    -e COLORED_OUTPUT="1"
+    -e COLORED_OUTPUT="${COLORED_OUTPUT_SUPPORTED}"
     -e DISPLAY="${DISPLAY}"
     -e CONTAINER_TOOL="${CONTAINER_TOOL}"
     -e ZWIFT_UID="${ZWIFT_UID}"
--- a/src/zwift.sh
+++ b/src/zwift.sh
@@ -11,6 +11,7 @@ readonly ZWIFT_DOCS="${WINE_USER_HOME}/AppData/Local/Zwift"
 readonly ZWIFT_DOCS_OLD="${WINE_USER_HOME}/Documents/Zwift" # TODO remove when no longer needed (301)
 
 if [[ -t 1 ]]; then
+    readonly COLORED_OUTPUT_SUPPORTED="1"
     readonly COLOR_WHITE="\033[0;37m"
     readonly COLOR_RED="\033[0;31m"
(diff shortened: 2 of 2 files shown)


src/update_zwift.sh
```diff
@@ -1,149 +1,181 @@
...
+        return 1
+    fi
 
-    # install dotnet48 for zwift
-    winetricks -q dotnet48
+    local i
+    for i in $(seq 5 -1 1); do
+        msgbox info "Waiting for update to finish... (${i})"
+        sleep 1
+    done
```
glennvl, line 105: There was already a sleep of 5 seconds here in the original script, so I kept it.

While testing I tried without the sleep and everything worked just fine. So I don't think it's actually needed. Why was the sleep there in the first place?

netbrain: Seem to have a vague memory that the install could get flakey if we didn't have this sleep.  But not entirely sure, as the code has changed quite a lot over the year with several contributions.

src/build-image.sh
```diff
@@ -1,43 +1,113 @@
...
+##### Basic configuration #####
 
+# Create array for container arguments
 declare -a container_args
 container_args=(
     -it
     --network bridge
     --name zwift
     --security-opt label=type:container_runtime_t
-    --device /dev/dri
```
netbrain, line 36: should this have been removed?

src/build-image.sh
```diff
@@ -1,43 +1,113 @@
...
     --hostname "${HOSTNAME}"
 
+    -e DEBUG="${DEBUG}"
+    -e COLORED_OUTPUT="${COLORED_OUTPUT_SUPPORTED}"
     -e DISPLAY="${DISPLAY}"
-    -e XAUTHORITY="${XAUTHORITY}"
```
netbrain, line 40: And this?

src/build-image.sh
glennvl: I moved things around a bit so they are the same as in the zwift launcher script.

Now the following chunk is exactly the same in build-image.sh (line 128-136) and zwift.sh (line 539-547):

```bash
# Check for proprietary nvidia driver and set correct device to use
if [[ -f "/proc/driver/nvidia/version" ]]; then
    if [[ ${CONTAINER_TOOL} == "podman" ]]; then
        container_args+=(--device="nvidia.com/gpu=all")
    else
        container_args+=(--gpus="all")
    fi
else
    container_args+=(--device="/dev/dri:/dev/dri")
fi
```

src/build-image.sh
glennvl: It's possible that the XAUTHORITY environment variable is not set, which would cause the `set -u` flag to trigger. So I added a guard to it instead of always adding without checking.

```bash
# line 58: always initialize, to make set -u happy
readonly XAUTHORITY="${XAUTHORITY:-}"

# line 123: only add if not empty
if [[ -n ${XAUTHORITY} ]]; then
    container_args+=(-e XAUTHORITY="${XAUTHORITY}")
fi
```

PR #310, opened 2026-03-05
Add window manager sanity checks and only invoke xhost if needed

glennvl:
I am trying to update wine to the latest version. The latest wine is causing permission issues with egl and the shm when x11 is used (it works with wayland). So I went diving into the documentation and found some potential issues and inefficiencies in the way we configure the window manager. Sadly they don't fix the egl problem, but I hope they may help others who have reported zwift not launching or black screen problems.

## Summary

- Add window manager sanity checks (messages and error exit)
- Only invoke xhost if needed (XAUTHORITY not set)
- Invoke xhost without hostname, before launching the container
- Invoke xhost in the build script if XAUTHORITY is not set

### Add window manager sanity checks

- ~Rely solely on XDG_SESSION_TYPE to detect the window manager~
  - ~According to the documentation it always exists and its value is one of: `wayland`, `x11`, `tty`~
  - ~The WAYLAND_DISPLAY is not set if XDG_SESSION_TYPE is not wayland, so the check does nothing~
  - ~The DISPLAY environment variable is always set. So if XDG_SESSION_TYPE is tty, it would detect x11.~
  - ~If the XDG_SESSION_TYPE is tty, exit with an error, we won't be able to launch zwift without window manager~
  - ~If the user wants to force anything here, he can launch zwift with `XDG_SESSION_TYPE=... zwift`~
- Check if all required environment variables are set and directories exist. If not, show an error message
  - Wayland: The XDG_RUNTIME_DIR and WAYLAND_DISPLAY environment variables are required
  - X11: The DISPLAY environment variable and /tmp/.X11_unix directory are required

### Only invoke xhost if needed

- Allowing xhost network access poses security risks (other users can read keystrokes etc)
- X11 provides an alternative way to allow access, the XAuthority cookie. We already use this (XAUTHORITY environment variable and volume). But we still invoke xhost regardless.
- The comments in the code used to say that invoking xhost is required when ZWIFT_UID is different from  UID. I tested this and it is not the case, the XAuthority cookie also works with different user/group ids.
- So, only invoke xhost if there is no XAuthority cookie (XAUTHORITY environment variable is empty)

### Invoke xhost without hostname

We used to invoke xhost with `xhost +local:$HOSTNAME`. But according to the xhost documentation, when using the `+local:`, anything after `:` is ignored. So Calling `xhost +local:` is equivalent to calling `xhost +local:$HOSTNAME` and equivalent to `xhost +local:whatever_i_want_to_write_here`.

What this means is that we don't need a hostname to run xhost. So we don't need to wait for the container to launch to get its id. Instead we can just invoke xhost before launching the container. This makes the launch procedure a lot simpler since we don't need to worry about interactive mode, etc.

## Discussion

I added `exit 1` for critical errors. I did some testing and I think there's no point in continuing if we know zwift will not launch. In my opinion it would only make it harder to pinpoint what exactly went wrong.

Examples are when the XDG_SESSION_TYPE is not set or tty or when any other environment variables or directories are missing. I also tested what happens if invoking xhost fails if there is not XAuthority cookie and it will cause an error in the zwift launcher and a zwift crash with a stacktrace (not even a black screen). So I think it's better we just exit and print the reason instead of letting other commands spam more output and eventually crashing anyway.

### DIFF
--- a/src/zwift.sh
+++ b/src/zwift.sh
@@ -624,6 +624,7 @@ fi
 # Create a volume if not already exists, this is done now as
 # if left to the run command the directory can get the wrong permissions
 if [[ ${CONTAINER_TOOL} == "podman" ]] && ! ${CONTAINER_TOOL} volume ls | grep -q "zwift-${USER}"; then
+    mshgbox info "Creating ${CONTAINER_TOOL} volume zwift-${USER}"
     if ${CONTAINER_TOOL} volume create "zwift-${USER}"; then
         msgbox ok "Created volume zwift-${USER}"
     else
@@ -634,10 +635,16 @@ fi
 
 # Only write environment variables to file when needed
 msgbox info "Writing environment variables to temporary file"
-printf '%s\n' "${container_env_vars[@]}" > "${container_env_file}"
+if printf '%s\n' "${container_env_vars[@]}" > "${container_env_file}"; then
+    msgbox ok "Environment variables written to temporary file"
+else
+    msgbox error "Failed to write environment variables to temporary file"
+    exit 1
+fi
 
 # Use xhost to allow container to access X11 if needed
 if [[ ${xhost_access_required} -eq 1 ]]; then
+    msgbox info "Invoking xhost"
     if command_exists xhost && xhost +local: > /dev/null; then
         msgbox ok "Container X11 access provided through xhost"
     else


perrin4869:
please note that `XDG_SESSION_TYPE` can be set to `tty` and still be running `x11`. This is the case if you boot into tty and run `startx`

glennvl:
@perrin4869 

> please note that `XDG_SESSION_TYPE` can be set to `tty` and still be running `x11`. This is the case if you boot into tty and run `startx`

Thanks for pointing that out. How do you think we should handle such cases?

- Is this also possible for wayland?
- Try to detect it? (ie `[[ ${XDG_SESSION_TYPE} == "tty" ]] && pgrep Xorg > /dev/null 2>&1`) Would that even be possible?
- Add extra info to the error message to point out that he can launch with `XDG_SESSION_TYPE=x11 zwift`?
- Just assume x11 if it's not wayland?
- ...

I'm probably leaning towards telling the user to manually set XDG_SESSION_TYPE, but I'm open to any solution.

perrin4869:
I think this discussion popped up a few years ago, I think `XDG_SESSION_TYPE` shouldn't be consulted because it's not reliable. Probably best to try to detect wayland via `WAYLAND_DISPLAY`, fallback to `DISPLAY`? Unfortunately I never ran Wayland so I can't help there

glennvl:
> I think this discussion popped up a few years ago, I think `XDG_SESSION_TYPE` shouldn't be consulted because it's not reliable. Probably best to try to detect wayland via `WAYLAND_DISPLAY`, fallback to `DISPLAY`? Unfortunately I never ran Wayland so I can't help there

I think in that case the simplest possible detection would be something like this:

```bash
if [[ -n ${WAYLAND_DISPLAY} ]]; then
    if [[ ${WINE_EXPERIMENTAL_WAYLAND} -eq 1 ]]; then
        window_manager="Wayland"
    else
        window_manager="XWayland"
    fi
else
    window_manager="XOrg"
fi
```

And then just rely on the further code to report errors if the required environment variables and directories are not present.

Or taking into account

> You can apparently detect X11 more reliably using [ -S "/tmp/.X11-unix/X${DISPLAY#*:}" ] && echo "X11 is running", it works in my system

```bash
if [[ -n ${WAYLAND_DISPLAY} ]]; then
    if [[ ${WINE_EXPERIMENTAL_WAYLAND} -eq 1 ]]; then
        window_manager="Wayland"
    else
        window_manager="XWayland"
    fi
elif [[ -n ${DISPLAY} ]] && [[ -S /tmp/.X11-unix/X${DISPLAY#*:} ]]; then
    window_manager="XOrg"
else
    # error
fi
```

perrin4869:
You can apparently detect X11 more reliably using `[ -S "/tmp/.X11-unix/X${DISPLAY#*:}" ] && echo "X11 is running"`, it works in my system

would probably be wise to test it on other systems/distros to make sure, but I think that's the way 😅

glennvl:
@perrin4869 @netbrain 

I've been doing some more research on how to detect wayland and x11. There's only a few ways described on the internet and none of them appear to be reliable. I'm seeing stuff that worked a year ago that doesn't work today.

Instead of trying to do something that is at best unreliable, I see two options:

- Keep the change I made in this PR, relying purely on XDG_SESSION_TYPE. But if it is equal to tty, inform the user that he can set XDG_SESSION_TYPE to wayland or x11 in the config file or on the zwift command line if there should've been a window manager.
- Don't try to detect anything. Just use WINE_EXPERIMENTAL_WAYLAND to decide between wayland or x11. And try to fallback to x11 if wayland was selected but XDG_RUNTIME_DIR or WAYLAND_DISPLAY is not set.

Or just use WAYLAND_DISPLAY to detect wayland and `-S /tmp/.X11-unix/X${DISPLAY#*:}` to detect x11 and hope for the best. I implemented this in 18ed759db201ee9bdab6f879341ee72e271aa850.

What do you think?

netbrain:
Latest approach with WAYLAND_DISPLAY to X11 socket fallback looks solid. Much more reliable than XDG_SESSION_TYPE.

netbrain:
Is this ready for review?

glennvl:
> Is this ready for review?

Yup, ready for review. But I'll have to rebase tomorrow since there's conflicts with the PR we just merged.

netbrain:
LGTM

PR #314, opened 2026-03-12
Optimize Docker startup by avoiding unnecessary recursive chown

### LINKED ISSUE #311: Loading takes a longgggg time!
## Describe the bug

Recently (since the update that added more to the script in terms of writing password to podman store and transferring variables to temp file) I;ve noticed that the script takes a long time to load zwift (approx 3-4 mins as opposed to the usual 1-2 I was seeing before)

This is to get to the initial bit  where it stops running as a script and starts the gui

Seems to get stuck on the writing environment variables bit

Almost seems like something finally errors/quits/times out and then everything kicks back in and loads as it should.

No podman updates/docker pulls or anything happening  and happens every time for me

Let me know what else you need from me and I'll get it sorted!

netbrain:
## Summary
- Skip `chown` entirely when the top-level directory already has correct ownership, making normal repeated launches near-instant
- Replace `chown -R` with selective `find`-based chown that only touches files with incorrect ownership
- Add progress message during slow ownership updates so users know what's happening
- Update slow-start troubleshooting docs

Closes #311

### DIFF
--- a/src/entrypoint.sh
+++ b/src/entrypoint.sh
@@ -129,8 +129,7 @@ if [[ ${CONTAINER_TOOL} == "docker" ]]; then
         # This avoids a costly recursive find on every normal startup
         local target="${1:?}"
         local result
-        result="$(find "${target}" -maxdepth 0 \( ! -user user -o ! -group user \) -print 2> /dev/null)" || true
-        [[ -d ${target} ]] && [[ -n ${result} ]]
+        [[ -d ${target} ]] && result="$(find "${target}" -maxdepth 0 \( ! -user user -o ! -group user \) -print 2> /dev/null)" && [[ -n ${result} ]]
     }
 
     update_ownership() {


netbrain:
@glennvl ready for review.

src/entrypoint.sh
```diff
@@ -124,12 +124,31 @@ if [[ ${CONTAINER_TOOL} == "docker" ]]; then
         sed -i "s/1000/${user_uid}/g" /etc/pulse/client.conf || return 1
     }
 
+    ownership_needs_update() {
+        # Quick check: if the top-level directory is already owned by user:user, assume everything is fine
+        # This avoids a costly recursive find on every normal startup
+        local target="${1:?}"
+        local result
+        result="$(find "${target}" -maxdepth 0 \( ! -user user -o ! -group user \) -print 2> /dev/null)" || true
+        [[ -d ${target} ]] && [[ -n ${result} ]]
```
glennvl:
```suggestion
        [[ -d ${target} ]] && result="$(find "${target}" -maxdepth 0 \( ! -user user -o ! -group user \) -print 2> /dev/null)" && [[ -n ${result} ]]
```

Could be a bit safer for error handling?

- Only execute find if it's a directory
- Only check contents of result of find succeeds

glennvl (commented):
I'm not entirely sure just checking the top level directory is enough. For the volume itself, sure, but we also give the user the option to mount a whole list of additional directories through configuration parameters (think things such as the logs directory, screenshots, ..., even graphics profiles).

Those additional directories may have incorrect ownership. Although I guess that could be considered user error.

- Should we actually even allow the user to provide a ZWIFT_UID and ZWIFT_GID? Should they not always be set to `id -u` and `id -g` respectively? The user could just switch accounts himself or run as a different user with `sudo -i other-user zwift`.
- Going even further, what was the reasoning behind trying to run everything as user inside the container and map that user to the actual user account?

glennvl:
> Thanks for the review!

In think this is a good start to simplify the switching uid/gid stuff.

> **Top-level check and bind-mounted dirs:** You're right that bind-mounted subdirectories (workouts, logs, screenshots) could have different ownership. I think this falls under user responsibility though — those are host paths the user explicitly provides, and chowning host files from inside the container would be surprising/destructive. The top-level check targets the volume-managed data where the container is responsible for ownership. Open to adding a note in the docs if you think that's worth calling out.

You are completely right that chowning them is surprising. That is how it currently works. I've run into it multiple times while testing with docker and different uid/gid. When running `CONTAINER_TOOL=docker ZWIFT_UID=2000 ZWIFT_GID=2000 zwift`, it changes ownership of the mounted files and directories on my pc, which I then have to manually fix again after testing to run zwift normally.

> **ZWIFT_UID/ZWIFT_GID:** Fair question. The main use case is multi-user setups where files need specific uid/gid ownership. `sudo -i other-user zwift` requires that user to exist with a full session which is more friction. That said, the volume variant in PR 316 makes this largely moot since ownership persists in the volume. I'd lean toward keeping them for backwards compatibility but open to deprecating in a future PR.

In my opinion allowing a uid/gid different from the user uid/gid is just too fragile. Sure, using `sudo -i other-user zwift` instead has some extra requirements, but it also just works. I'd like to see at the very least a big fat warning (even an error msgbox) when the user does this, saying if anything does not work it's his own responsibility.

> **Why run as user inside the container:** Docker containers run as root by default. Wine shouldn't run as root (security + wine warnings), and files written to bind-mounted host directories (activities, logs, screenshots) need correct ownership — otherwise they'd be owned by root on the host. The entrypoint does `gosu user:user` to drop privileges after setup. Podman avoids this entirely with `--userns keep-id`.

I think there's some improvements that can be made here to simplify things. I'll get back to this soon. :wink: 

> **Inline suggestion:** Good call on the defensive ordering in `ownership_needs_update()`, applied.

:+1:

Issue #315, opened 2026-03-12
Volume-based Zwift installation to eliminate per-launch chown


netbrain:
## Problem

When using Docker with a uid/gid that differs from the container default (1000:1000), the entrypoint runs chown on every launch. The container is ephemeral, so /home/user is recreated from the image layer each time with 1000:1000 ownership. Currently only ZWIFT_DOCS is persisted as a volume.

## Suggested solution

Decouple the Zwift installation from the container image. Instead of baking Zwift into the image, offer a volume-based installation mode:

- **Minimal container image**: only wine + dependencies, no Zwift pre-installed
- **First run**: user runs `zwift --install`, which mounts an empty volume at /home/user and runs the Zwift installer into it
- **Subsequent runs**: mount the same volume, launch directly. No chown needed since ownership persists in the volume
- **Updates**: Zwift's own updater handles it, or user runs `zwift --update`

This approach:
- Eliminates the per-launch chown entirely
- Shrinks the container image (no bundled Zwift installation)
- Mirrors how Zwift works on a real machine (install once, self-update)
- Could coexist with the current pre-installed approach as an opt-in alternative

This aligns with the second variant described in issue #28  (minimal container with volume-based Zwift download on first run).

netbrain:
Related to #311

glennvl:
Let me start out by saying I can see you already put a lot of thought and effort into this. And I appreciate your effort and that you created this great project that allows us to easily run Zwift on Linux! 

## My humble opinion

What we have now is a simple and elegant solution:

- A container with Zwift installed that just works
- Zwift updates are handled server side, the end user does not have to worry about anything

Are we sure adding (potentially more fragile) variants fits within the philosophy of this project?

- Letting the user handle Zwift installation and updates moves unnecessary responsibility to the end user
- The other variant that uses native wine especially sounds like an entirely different project to me

## Issues with doing installs and updates on the end-user's setup

The launcher does not work to actually start Zwift. But we still need it to update Zwift. So when installing or updating Zwift we have to wait for the launcher to finish the update, then start Zwift ourselves. We do so by waiting until the local Zwift version is the same as the version in the <https://cdn.zwift.com/gameassets/Zwift_Updates_Root/Zwift_ver_cur.xml> file (referred to as Zwift_ver_cur.xml file from here on).

When updating Zwift automatically server-side using a github action, this is a great solution! But only because we only trigger the update only after checking the value of that same Zwift_ver_cur.xml file.

[...]

PR #316, opened 2026-03-12
Add volume variant for persistent /home/user mount

netbrain:
## Summary

- Adds `ZWIFT_VARIANT=volume` mode where `/home/user` is mounted as a persistent Docker volume (`zwift-home-$USER`)
- On first run, the container runtime auto-populates the volume from the image
- Ownership persists in the volume, skipping per-launch `chown` for Docker users with uid/gid != 1000
- Pre-launch sync: when the image is updated, an init container rsyncs changes into the volume while preserving user data
- Adds `zwift-volume` Nix package and `variant` option to NixOS module
- Adds documentation: dedicated volume variant page, updated config options, NixOS docs, and slow-start troubleshooting

## How it works

Docker and Podman automatically copy image contents into a new named volume on first mount. By mounting `/home/user` as a volume instead of just the Zwift docs directory, the entire wine prefix persists between runs.

**Volume sync on image update:** Before launching, `zwift.sh` compares the image's OCI version label against a marker file in the volume. If they differ, it runs a short-lived container that mounts the volume at `/mnt/volume` (keeping the image's `/home/user` visible) and uses `rsync` to copy updated files into the volume. User data (activities, workouts, logs, preferences, screenshots) is excluded from the sync.

### Files changed

| File | Change |
| --- | --- |
| `src/zwift.sh` | Volume mount switching, pre-launch version check and rsync sync |
| `src/entrypoint.sh` | Skip chown block when `ZWIFT_VOLUME=1` |
| `src/update_zwift.sh` | Guard `rm -rf $ZWIFT_DOCS` behind `ZWIFT_VOLUME` check |
| `src/Dockerfile` | Add `rsync` package |
| `.github/workflows/zwift_build_from_scratch.yaml` | Embed `ZWIFT_IMAGE_VERSION` env var at commit time |
| `flake.nix` | Add `variant` option, `zwift-volume` package |
| `docs/advanced/volume-variant.md` | New documentation page |
| `docs/configuration/options.md` | Add `ZWIFT_VARIANT` to options table |
| `docs/advanced/nixos.md` | Add `variant` to NixOS config example |
| `docs/troubleshooting/slow-start.md` | Mention volume variant as a solution |

## Test plan

- [x] First run: verify volume is populated and chown is skipped
- [x] Second run: verify it launches directly, volume marked up to date
- [x] Simulate image update: change version label, verify rsync triggers
- [x] Default variant (`container`) still works unchanged

<details>
<summary>Local testing instructions (no Zwift installation required)</summary>

[...]

PR #333, opened 2026-03-16
Refactor uid gid

glennvl:
Related to #315.

> **Spoiler**: It's not just 25 lines of code.

## Summary

This PR refactors how we deal with UIDs and GIDs.

Current implementation:

- Podman: Can automatically remap ids, and we use this feature
- Docker: Does not support remapping ids out of the box, container starts as root, then reown and switch user

New implementation:

- Podman: No change
- Docker:
  - Start container as user instead of root (user also no longer has root privileges)
  - Entrypoint no longer has root access, don't reown anything and no need to switch user
  - In the zwift.sh script: Remap the container UID/GID using a temporary dockerfile if needed
  - In the zwift.sh script: Remap the volume UID/GID if needed

> **Important**: This change is NOT compatible with the end user requesting Zwift to run with different ids than the ids of $USER. So support for the ZWIFT_UID and ZWIFT_GID configuration options has been removed. A real user is required, use `sudo -u other_user zwift` instead of `ZWIFT_UID=other_user_uid ZWIFT_GID=other_user_gid zwift`.

### Exit with error when ZWIFT_UID or ZWIFT_GID are set

We now remap the container user to the host user. This does not work with custom ids.

If the end user has a ZWIFT_UID or ZWIFT_GID set, report an error that they are no longer supported and exit the zwift script. The end user has to remove these configuration options to be able to continue using zwift.

### Building the image locally using build-image.sh

The build script is fragile when the host user has a UID/GID that is not equal to 1000:1000. For example the XAUTHORITY cookie may not work, resulting in the installer not having access to X11. This problem is exagerated if the end user manually sets a ZWIFT_UID/ZWIFT_GID. Currently the only safe way to use the build script is with a host user that has UID/GID 1000:1000.

Since we know the UID/GID in advance, we might as well take advantage of it and pass them as build argument to the Dockerfile. We immediately create the container user with the correct UID/GID. This in combination with no longer allowing the end user to specify a ZWIFT_UID/ZWIFT_GID makes it safe to use the build script on a local user with UID/GID not equal to 1000:1000.

> **Important**: When docker is used as container tool, we can therefore no longer assume that the UID/GID of container user is 1000:1000.

> **Important**: No change when the container tool is podman. The container user ALWAYS has UID/GID 1000:1000.

### Don't launch container as root, use user instead

The following changes were made to the container user:

- Don't give root access to the container user, it's not needed
- Use a USER directive to automatically switch to user instead of starting as root

> **Advantage**: No root access inside the container, making privilege escalation less likely.

> **Note**: If root access to the container is required, the end user can still do so by using `docker exec --user root`.

```Dockerfile
# Build script: Allow specifying the UID and GID to use
ARG USER_UID=1000
ARG USER_GID=1000

# Create user with UID and GID, but NO root access
RUN addgroup --gid ${USER_GID} user \
 && adduser --uid ${USER_UID} --gid ${USER_GID} --disabled-password --comment '' user

# Switch to user (starts container as user with both docker and podman)
USER user
```

> **Important**: The entrypoint script is now launched as user and no longer has root privileges. It can no longer be used to change ownership.

### Changing ownership: Remapping the container user to the host user

Since the entrypoint script now runs as user instead of root, it can't perform a chown. We now remap the container to the host user in the zwift script instead.

The remapping is done in two stages.

#### 1. Remap the container user to the host user by changing its UID/GID to be the same as the host user UID/GID.

We perform this remapping using a simple dockerfile that starts from the requested IMAGE and VERSION (default: docker.io/netbrain/zwift). The dockerfile temporarily switches the USER to root to execute the same commands that used to be executed in the entrypoint script. These commands change the container user UID/GID to be the same as the host user UID/GID. Before switching the USER back to user. The dockerfile is generated on the fly when needed.

A label is also attached with the repository image digest of the requested zwift image IMAGE:VERSION. The remapping only has to be done if the image digest changed (IMAGE:VERSION changed).

```Dockerfile
# user_uid = host user uid
# user_gid = host user gid
# image_digest = ${IMAGE}:${VERSION} repository digest
   
FROM ${IMAGE}:${VERSION}
USER root
RUN usermod -ou ${user_uid} user \
 && groupmod -og ${user_gid} user \
 && mkdir -p /run/user/${user_uid} \
 && chown -R user:user /run/user/${user_uid} \
 && sed -i "s|/run/user/1000|/run/user/${user_uid}|g" /etc/pulse/client.conf
USER user
ENTRYPOINT ["entrypoint"]
LABEL org.opencontainers.image.base.digest="${image_digest}"
```
> **Question**: The `mkdir -p /run/user/${user_uid}` and `chown -R user:user /run/user/${user_uid}` were also done in the entrypoint. But are the actually needed?

The freshly built image is tagged as `netbrain/zwift:remapped_user_$UID_$GID` and used to launch zwift instead of IMAGE:VERSION. Including the UID/GID in the tag name is important for several reasons:

- So we know if an image with the correct UID/GID already exists. Otherwise we would always have to rebuild since we cannot know if the end user launched as a different user this time or if he changed the UID/GID of his account.
- If the end user wants to be able to use zwift on multiple users, we don't want to constantly overwrite the tag. This would also break things if the end user wants to launch multiple zwift instances at the same time.

> **Important**: The remapping also has to be done for host users with UID/GID 1000:1000 since we can no longer assume that the container user was created with UID/GID 1000:1000. The performance overhead is negligible (remapping is super fast if ids have not changed).

> **Important**: If the user uses a local IMAGE:VERSION (instead of one that is published on docker hub or another registry), there is no repository digest. So the remapping will be done each launch (if nothing changed, the overhead is negligibly). This, for example, is the case when building the image locally using the build script.

#### 2. Make sure the ownership of the zwift-$USER volume is correct

We also need to make sure the UID/GID of the zwift documents directory, mounted from the zwift-$USER volume, is correct. To check its UID/GID, we mount it in a temporary container. If the UID/GID turns out to be wrong, another temporary container is launched, this time with root user, and a recursive chown is performed.

- Check if the zwift-$USER volume UID/GID is correct:

  ```bash
  docker run \
      --rm \
      -it \
      -v zwift-$USER:/zwift-docs \
      --entrypoint bash \
      netbrain/zwift:remapped_user_$UID_$GID \
      -c "[[ ! -O /zwift-docs ]] || [[ ! -G /zwift-docs ]]"
  ```

- Update the zwift-$USER volume UID/GID:

  ```bash
  docker run \
      --rm \
      --user root \
      -it \
      -v zwift-$USER:/zwift-docs \
      --entrypoint bash \
      netbrain/zwift:remapped_user_$UID_$GID \
      -c "chown -R $UID:$GID /zwift-docs"
  ```

The check whether the UID/GID of the volume are correct is always performed. There are two situations where remapping is actually required:

- The zwift-$USER volume is initially created as root. This is the case even if the container is launched as user. Se we cannot rely on the run command to automatically create the volume correctly. Instead we manually create the volume if it does not already exists (we already did this for podman) and then update the UID/GID.
- If the end user changes his UID/GID, the ids in the container and volume will no longer be correct. The dockerfile will execute to remap the container user and we also have to chown the volume.

> **Note**: The run_zwift script now also checks the UID/GID of the zwift documents directory. If they are incorrect it reports an error and exits.

## Other changes

Just two minor changes I smuggled into this PR:

* Reformat the default graphics profile a bit to make it more readable.
* When DRYRUN is not set, print the environment variables and command as debug messages.

### DIFF
--- a/src/zwift.sh
+++ b/src/zwift.sh
@@ -364,13 +364,13 @@ else
         local tag_name="${1:?}"
 
         local latest_image_digest=""
-        if ! latest_image_digest="$(${CONTAINER_TOOL} inspect "${IMAGE}:${VERSION}" --format '{{.Digest}}')"; then
-            msgbox error "Failed to get ${IMAGE}:${VERSION} image digest"
-            exit 1
+        if ! latest_image_digest="$(${CONTAINER_TOOL} inspect "${IMAGE}:${VERSION}" --format '{{index .RepoDigests 0}}' 2> /dev/null)"; then
+            msgbox warning "Failed to get ${IMAGE}:${VERSION} image digest"
+            return 0
         fi
 
         local current_image_digest=""
-        if ! current_image_digest="$(${CONTAINER_TOOL} inspect "${tag_name}" --format '{{index .Config.Labels "org.opencontainers.image.base.digest"}}')"; then
+        if ! current_image_digest="$(${CONTAINER_TOOL} inspect "${tag_name}" --format '{{index .Config.Labels "org.opencontainers.image.base.digest"}}' 2> /dev/null))"; then
             msgbox info "Failed to get ${tag_name} base image digest, may not exist yet"
             return 0
         fi
@@ -389,9 +389,9 @@ else
         fi
 
         local image_digest=""
-        if ! image_digest="$(${CONTAINER_TOOL} inspect "${IMAGE}:${VERSION}" --format '{{.Digest}}')"; then
-            msgbox error "Failed to get ${IMAGE}:${VERSION} image digest"
-            return 1
+        if ! image_digest="$(${CONTAINER_TOOL} inspect "${IMAGE}:${VERSION}" --format '{{index .RepoDigests 0}}' 2> /dev/null)"; then
+            msgbox warning "Failed to get ${IMAGE}:${VERSION} image digest"
+            image_digest="${IMAGE}:${VERSION}"
         fi
 
         {
--- a/src/zwift.sh
+++ b/src/zwift.sh
@@ -422,22 +422,20 @@ else
         fi
     }
 
+    msgbox info "Remapping container user to host user"
     container_uid="${host_uid}"
     container_gid="${host_gid}"
-    if [[ ${host_uid} -ne 1000 ]] || [[ ${host_gid} -ne 1000 ]]; then
-        msgbox info "Remapping container user to host user"
-        container_image="netbrain/zwift"
-        container_image_version="remapped_user_${container_uid}_${container_gid}"
-        if remap_build_required "${container_image}:${container_image_version}"; then
-            if remap_dockerfile="$(create_remap_dockerfile "${container_uid}" "${container_gid}")" && build_remap_dockerfile "${container_image}:${container_image_version}" "${remap_dockerfile}"; then
-                msgbox ok "Remapped container user to host user"
-            else
-                msgbox error "Failed to remap container user to host user"
-                exit 1
-            fi
+    container_image="netbrain/zwift"
+    container_image_version="remapped_user_${container_uid}_${container_gid}"
+    if remap_build_required "${container_image}:${container_image_version}"; then
+        if remap_dockerfile="$(create_remap_dockerfile "${container_uid}" "${container_gid}")" && build_remap_dockerfile "${container_image}:${container_image_version}" "${remap_dockerfile}"; then
+            msgbox ok "Remapped container user to host user"
         else
-            msgbox ok "${container_image}:${container_image_version} is up to date"
+            msgbox error "Failed to remap container user to host user"
+            exit 1
         fi
+    else
+        msgbox ok "${container_image}:${container_image_version} is up to date"
     fi
 fi
 
--- a/src/zwift.sh
+++ b/src/zwift.sh
@@ -175,10 +175,17 @@ readonly ZWIFT_FG="${ZWIFT_FG:-0}"
 readonly ZWIFT_NO_GAMEMODE="${ZWIFT_NO_GAMEMODE:-0}"
 readonly WINE_EXPERIMENTAL_WAYLAND="${WINE_EXPERIMENTAL_WAYLAND:-0}"
 readonly NETWORKING="${NETWORKING:-bridge}"
-readonly ZWIFT_UID="${ZWIFT_UID:-${UID}}"
-readonly ZWIFT_GID="${ZWIFT_GID:-$(id -g)}"
 readonly VGA_DEVICE_FLAG="${VGA_DEVICE_FLAG:-}"
 readonly PRIVILEGED_CONTAINER="${PRIVILEGED_CONTAINER:-0}"
+
+# No longer supported configuration environment variables
+readonly ZWIFT_UID="${ZWIFT_UID:-}"
+readonly ZWIFT_GID="${ZWIFT_GID:-}"
+if [[ -n ${ZWIFT_UID} ]] || [[ -n ${ZWIFT_GID} ]]; then
+    msgbox error "ZWIFT_UID and ZWIFT_GID options are no longer supported"
+    msgbox error "  To start Zwift as a different user, use: sudo -i username zwift"
+    exit 1
+fi
 
 # Initialize CONTAINER_TOOL: Use podman if available
 msgbox info "Looking for container tool"
@@ -206,8 +213,8 @@ declare -a parameters_to_print
 parameters_to_print=(
     DEBUG VERBOSITY CONTAINER_TOOL IMAGE VERSION SCRIPT_VERSION DONT_CHECK DONT_PULL DONT_CLEAN DRYRUN INTERACTIVE
     CONTAINER_EXTRA_ARGS ZWIFT_USERNAME ZWIFT_PASSWORD ZWIFT_WORKOUT_DIR ZWIFT_ACTIVITY_DIR ZWIFT_LOG_DIR ZWIFT_SCREENSHOTS_DIR
-    ZWIFT_OVERRIDE_GRAPHICS ZWIFT_OVERRIDE_RESOLUTION ZWIFT_FG ZWIFT_NO_GAMEMODE WINE_EXPERIMENTAL_WAYLAND NETWORKING ZWIFT_UID
-    ZWIFT_GID VGA_DEVICE_FLAG PRIVILEGED_CONTAINER DBUS_SESSION_BUS_ADDRESS DISPLAY WAYLAND_DISPLAY XAUTHORITY XDG_RUNTIME_DIR
+    ZWIFT_OVERRIDE_GRAPHICS ZWIFT_OVERRIDE_RESOLUTION ZWIFT_FG ZWIFT_NO_GAMEMODE WINE_EXPERIMENTAL_WAYLAND NETWORKING
+    VGA_DEVICE_FLAG PRIVILEGED_CONTAINER DBUS_SESSION_BUS_ADDRESS DISPLAY WAYLAND_DISPLAY XAUTHORITY XDG_RUNTIME_DIR
 )
 for parameter_to_print in "${parameters_to_print[@]}"; do
     parameter_print_value="$(declare -p "${parameter_to_print}")"
@@ -342,26 +349,24 @@ container_args=()
 declare -a entrypoint_args
 entrypoint_args=()
 
+# Initialize user ids
+host_uid="${UID}"
+host_gid="$(id -g)"
 if [[ ${CONTAINER_TOOL} == "podman" ]]; then
-    # Podman has to use container id 1000
-    # Local user is mapped to the container id
-    local_uid="${ZWIFT_UID}"
     container_uid=1000
-    container_gid=1000
-    container_args+=(--userns "keep-id:uid=${container_uid},gid=${container_gid}")
+    container_args+=(--userns="keep-id:uid=1000,gid=1000")
 else
-    # Docker will run as the id's provided.
-    local_uid="${UID}"
-    container_uid="${ZWIFT_UID}"
-    container_gid="${ZWIFT_GID}"
+    container_uid="${host_uid}"
+    container_env_vars+=(
+        HOST_UID="${host_uid}"
+        HOST_GID="${host_gid}"
+    )
 fi
 
 # Define base container environment variables
 container_env_vars+=(
     DEBUG="${DEBUG}"
     VERBOSITY="${VERBOSITY}"
-    ZWIFT_UID="${container_uid}"
-    ZWIFT_GID="${container_gid}"
     CONTAINER_TOOL="${CONTAINER_TOOL}"
 )
 
@@ -587,18 +592,13 @@ fi
 if [[ ${window_manager} == "Wayland" ]]; then
     msgbox info "Using Wayland window manager"
 
-    if [[ ${ZWIFT_UID} -ne ${UID} ]]; then
-        msgbox error "Wayland does not support ZWIFT_UID different to your id of ${UID}"
-        exit 1
-    fi
-
     if [[ -n ${XDG_RUNTIME_DIR} ]] && [[ -n ${WAYLAND_DISPLAY} ]]; then
         container_env_vars+=(
-            XDG_RUNTIME_DIR="${XDG_RUNTIME_DIR//${local_uid}/${container_uid}}"
+            XDG_RUNTIME_DIR="${XDG_RUNTIME_DIR//${host_uid}/${container_uid}}"
             WAYLAND_DISPLAY="${WAYLAND_DISPLAY}"
             WINE_EXPERIMENTAL_WAYLAND="1"
         )
-        container_args+=(-v "${XDG_RUNTIME_DIR}/${WAYLAND_DISPLAY}:${XDG_RUNTIME_DIR//${local_uid}/${container_uid}}/${WAYLAND_DISPLAY}")
+        container_args+=(-v "${XDG_RUNTIME_DIR}/${WAYLAND_DISPLAY}:${XDG_RUNTIME_DIR//${host_uid}/${container_uid}}/${WAYLAND_DISPLAY}")
     else
         msgbox error "Required environment variables XDG_RUNTIME_DIR and/or WAYLAND_DISPLAY are not set"
         msgbox error "Falling back to XWayland" 5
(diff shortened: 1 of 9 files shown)


netbrain:
Down the rabbit hole we go! :rabbit:

src/zwift.sh
```diff
@@ -392,6 +380,156 @@ for arg; do
...
+    local repo_digest
+    repo_digest="$(${CONTAINER_TOOL} inspect "${tag_name}" --format '{{index .RepoDigests 0}}' 2> /dev/null)" || return 1
+    echo "${repo_digest}"
+}
+
+image_digest_label() {
+    local tag_name="${1:?}"
+
+    local digest_label
+    digest_label="$(${CONTAINER_TOOL} inspect "${tag_name}" --format '{{index .Config.Labels "org.opencontainers.image.base.digest"}}' 2> /dev/null))" || return 1
```
netbrain: Syntax bug: extra `)` in the command substitution.

```bash
digest_label="$(... 2> /dev/null))" || return 1
#                               ^^ two closing parens, should be one
```

This will either cause a parse error or silently break the function.

src/zwift.sh
```diff
@@ -392,6 +380,156 @@ for arg; do
...
+    local image_digest
+    image_digest="$(image_repo_digest "${IMAGE}:${VERSION}" || echo "Unknown")"
+
+    echo "FROM ${IMAGE}:${VERSION}"
+    echo "USER root"
+    echo "RUN usermod -ou ${user_uid} user \\"
+    echo " && groupmod -og ${user_gid} user \\"
+    echo " && mkdir -p /run/user/${user_uid} \\"
+    echo " && chown -R user:user /run/user/${user_uid} \\"
+    echo " && sed -i \"s|/run/user/1000|/run/user/${user_uid}|g\" /etc/pulse/client.conf"
```
netbrain: This hardcodes `/run/user/1000` as the source for the sed replacement. Works for the published image (always 1000:1000), but if someone builds locally with `build-image.sh` — which now passes `--build-arg USER_UID=${host_uid}` — the pulse config already has their UID baked in.

If they later run as a different user, the sed won't match and audio breaks silently. Consider capturing the actual current value from the image instead of assuming 1000.

src/zwift.sh
```diff
@@ -392,6 +380,156 @@ for arg; do
...
+    else
+        msgbox error "Failed to create volume zwift-${USER}"
+        exit 1
+    fi
+fi
+
+volume_remap_required() {
+    ${CONTAINER_TOOL} run --rm \
+        -v "zwift-${USER}:/zwift-docs" \
+        -it --entrypoint bash \
```
netbrain: The `-it` flags will fail in non-interactive contexts (cron, systemd services, piped scripts). These ephemeral containers don't need a TTY — consider dropping `-it` or at least the `-t`.

netbrain (commented):
Thanks for the thorough writeup and the work here. A few inline issues flagged below, plus some architectural thoughts.

I've been looking at this alongside PR 316 (volume variant), which solves the same non-1000 uid/gid pain point from a different angle — by persisting `/home/user` as a volume so ownership is written once and never needs per-launch chown or remapping.

The rootless container change here (USER directive, dropping gosu/sudo) is a genuine security improvement and I think it's valuable on its own. My concern is with the runtime UID remapping machinery — generating temp Dockerfiles, running `docker build` at launch, spinning up ephemeral containers for volume ownership checks. That's a lot of moving parts that can fail in ways that are hard to debug for end users, and it still has per-launch overhead.

I think the ideal path might be to split the rootless container change from the runtime remapping. The rootless Dockerfile changes could land independently, and the uid mismatch problem might be better solved by volume persistence than runtime remapping.

Thoughts?

src/zwift.sh
glennvl: Removed `-it` in a58ccb71262b8f7d9120320c8c6b8de7a6194694.

src/zwift.sh
glennvl: Fixed in 55d903df8979e7a97d377054028df986a314a3c0.

src/zwift.sh
glennvl: That's my fault for writing verbose code during testing and then only doing a smoke test after cleaning up the code. Fixed in ec9c69c17e4ca932c6ac17d36eb8c4d5369ebf13.

glennvl:
> Thanks for the thorough writeup and the work here. A few inline issues flagged below, plus some architectural thoughts.
> 
> I've been looking at this alongside PR 316 (volume variant), which solves the same non-1000 uid/gid pain point from a different angle — by persisting `/home/user` as a volume so ownership is written once and never needs per-launch chown or remapping.

In that sense the two implementations offer the same benefit. Remapping only needs tot be done once after each image update.

> The rootless container change here (USER directive, dropping gosu/sudo) is a genuine security improvement and I think it's valuable on its own. My concern is with the runtime UID remapping machinery

There's always going to be some sort of runtime UID remapping machinery. We don't know the UID/GID of the host user, so the container user has to be remapped to the host user some way or another.

> — generating temp Dockerfiles, running `docker build` at launch, spinning up ephemeral containers for volume ownership checks. That's a lot of moving parts that can fail in ways that are hard to debug for end users,

The remapping has to be done either in the entrypoint script or in the zwift script. Since this implementation runs the entrypoint script as user instead or root, the remapping can no longer be done in the entrypoint script.

In both this implementation and #316 ephemeral containers are used to check the version and to do the remapping if needed. The temporary dockerfile approach to simply update the container user UID/GID seems far less fragile to me than attempting to rsync the updated zwift install into a persistent volume. The dockerfile uses the same commands that were used in entrypoint before (should_change_user_ids part, which is still executed in the volume variant).

The second part of the remapping is to make sure the volume has the correct permissions. It is required for the two reasons mentioned in the OP. The container no longer starts as root, but the volume is still automatically created as root, so we need to reown it as user. And second it is possible that the user decides to change his account UID/GID, in that case the volume also has to be updated. If the user changes his UID/GID in the volume variant, it would trigger the change_user_ids funtion. Assuming the volume is owned by the current container user UID/GID, the usermod and groupmod would automatically update the volume.

Note that when changing the ownership inside entrypoint, it will also change ownership of other mounted directories and files which are mounted to subdirectories of the volume (whether it is /home/user or just AppData/Local/Zwift). These are files and directories on the host. The implementation in this PR does not have that problem (since it only mounts the volume itself when changing ownership).

Note that when running zwift with UID/GID not equal to 1000:1000, the volume variant still runs the change_user_ids code in the entrypoint script each time the container is launched. The implementation in this PR persists the UID/GID change into a new image, so it only has to run once (each time the original image is updated).

> and it still has per-launch overhead.

The per-launch overhead is neglibible.

Normal usage:

- first install -> remap using dockerfile (only slow if UID/GID != container) and create+own volume (fast)
- docker.io/netbrain:zwift changed -> remap using dockerfile (slow) and check volume (fast)
- docker.io/netbrain:zwift not changed -> check label (fast) and check volume (fast)

Special case, user changes UID/GID:

- remap using dockerfile (slow)
- change ownership of volume (medium)

Special case, user uses local image instead of image hosted on container registry:

- docker does not calculate a repo digest for local images (docker limitation, podman does)
- remap using dockerfile (fast because no actual changes are done)
- check volume (fast)

Comparing runtime overhead of this implementation vs the volume variant from #316:

- this implementation
  - check if container needs to be remapped: 2x docker inspect
  - check if volume needs to be remapped: 1x ephemeral container with trivial command
- volume variant
  - check if volume needs to be updated: 1x docker inspect + 1x ephemeral container with trivial command
  - if UID/GID not equal to 1000:1000: change_user_ids in entrypoint (usermod, groupmod)

> I think the ideal path might be to split the rootless container change from the runtime remapping. The rootless Dockerfile changes could land independently,

With a rootless container we can't do any remapping in the entrypoint script anymore. So we can't really separate those two.

> and the uid mismatch problem might be better solved by volume persistence than runtime remapping.

Well, to be frank: the volume variant is doing more runtime remapping than the implementation in this PR. :see_no_evil: 

- This PR: UID/GID remapping is persisted in an image
- Volume variant: Only ownership of files and directories is persisted. The user UID/GID still needs to be remapped each time the zwift container is launched.

> Thoughts?

Yes :sweat_smile:. I really don't think we should persist the entire user home in a volume. If anything changes to the directory structure (in the zwift install or elsewhere), this could easily break. The zwift install also isn't user local data, it's image data. I'm all for looking into adding a second volume for other user data if that can improve performance, but in my opinion the zwift install should always be part of the image.

A bit off topic, but I did a quick check to see where the shader cache is stored. On my system with NVIDIA graphics card, a `/home/user/.cache/nvidia/GLCache` directory exists. And on my system with AMD graphics card, two directories exist: `/home/user/.cache/mesa_shared_cache` and `/home/user/.cache/mesa_shader_cache_db`. So it could be as simple as adding a volume for `/home/user/.cache` if we want to persist the shader cache (if the same location is also used by other graphics cards).

src/zwift.sh
glennvl: I'm surprised this wasn't caught by either shfmt or shellcheck.

src/zwift.sh
```diff
@@ -175,11 +175,18 @@ readonly ZWIFT_FG="${ZWIFT_FG:-0}"
...
 readonly VGA_DEVICE_FLAG="${VGA_DEVICE_FLAG:-}"
 readonly PRIVILEGED_CONTAINER="${PRIVILEGED_CONTAINER:-0}"
 
+# No longer supported configuration environment variables
+readonly ZWIFT_UID="${ZWIFT_UID:-}"
+readonly ZWIFT_GID="${ZWIFT_GID:-}"
+if [[ -n ${ZWIFT_UID} ]] || [[ -n ${ZWIFT_GID} ]]; then
+    msgbox error "ZWIFT_UID and ZWIFT_GID options are no longer supported"
+    msgbox error "  To start Zwift as a different user, use: sudo -i username zwift"
+    exit 1
```
glennvl, line 196: We need to figure out something better here. Obviously it's not as simple as just running `sudo -u username zwift`. On XOrg that probably works after running `xhost +si:localuser:username`, but on Wayland it's much more involved.

Do we know of end users who actually use the ZWIFT_UID/ZWIFT_GID feature?
Can we get feedback from them?

src/zwift.sh
```diff
@@ -392,6 +380,158 @@ for arg; do
...
+if [[ ${CONTAINER_TOOL} == "podman" ]]; then
+    container_uid=1000
+    container_gid=1000
+
+    container_args+=(--userns="keep-id:uid=${container_uid},gid=${container_gid}")
+else
+    container_uid="${host_uid}"
+    container_gid="${host_gid}"
+
+    container_image="netbrain/zwift"
```
glennvl:
```suggestion
    container_image="${IMAGE#docker.io/}"
```

Maybe use the orignal image name instead of netbrain/zwift?

src/zwift.sh
```diff
@@ -392,6 +380,158 @@ for arg; do
...
+    container_image_version="remapped_user_${container_uid}_${container_gid}"
```
glennvl:
```suggestion
    container_image_version="uid_${container_uid}_gid_${container_gid}"
```

Maybe use a bit shorter tag name?

src/zwift.sh
netbrain: Anyone who has uid/gid other than 1000/1000 i assume. I have two accounts on my main zwift machine, my user account is 1000/1000, and my wife's is 1001:1001 (which means she has to wait a minute or two extra for the container to launch :smiling_imp:). 

But none of us have used zwift actively as of late due to other life responsibilities. :-1: 

Previously i simply had a single user account 1000/1000 and used USER=user1 zwift and USER=user2 zwift to switch between, which arguably worked better. 

I've been pondering if supporting multiple containerizations solution is the wrong way forward, maybe choose one? maybe that isn't docker or even podman, maybe its lxc container, or maybe even something totally different, like snaps, flatpak, appimage. I don't know...

On the flipside, some users have found docker to work, while having trouble with podman, and some probably the other way around.. So there's really no guarantee that we find a isolation technology that would be suitable for all, that was also my reason for having the #28  issue, to provide alternatives to those who don't fit the bill.

So yeah, down the rabbit hole we go :rabbit2:  :100: :sweat_smile:

glennvl: > Anyone who has uid/gid other than 1000/1000 i assume. I have two accounts on my main zwift machine, my user account is 1000/1000, and my wife's is 1001:1001 (which means she has to wait a minute or two extra for the container to launch 😈).

If she logged into her user account and started zwift from there she'd not be using the ZWIFT_UID/ZWIFT_GID feature. Since they would be automatically set to the ids of her account. Which is supported and streamlined by this PR.

With using the ZWIFT_UID/ZWIFT_GID feature I meant setting them to a different value than the logged in account. An example would be you and your wife zwifting at the same time on the same computer.

- You'd start your zwift instance normally (just run zwift)
- And start your wife's zwift instance as `ZWIFT_UID=1001 ZWIFT_GID=1001 zwift`

I think that always had issues since not everything would get mounted with the same ids and you'd need to manually update permissions for some files to make it work. Especially the XDG_RUNTIME_DIR and related stuff I think.

When starting two instances at the same time, indeed using the USER option is probably a more stable and reliable option.

glennvl: > I've been pondering if supporting multiple containerizations solution is the wrong way forward, maybe choose one? maybe that isn't docker or even podman, maybe its lxc container, or maybe even something totally different, like snaps, flatpak, appimage. I don't know...

I think we have a good solution going here with the zwift container image. Podman seems to be better at doing what we need out of the box with its support for remapping the user. But we got things working with docker, so we might as well keep supporting it.

The UID/GID implementation in this PR brings feature parity between docker and podman. I think it's the closest we can get to make the docker remapping behave the same as the automatic podman remapping.

It comes at the cost of allowing the user to specify a UID/GID different from the one of the linux account from which zwift is started. But that feature always behaved differently on docker and podman (did it even work on podman?) and was fragile with docker at best.

glennvl: > So yeah, down the rabbit hole we go 🐇 💯 😅

I think we should stick with the containerized solution we got and try to make that as good as we can. And I think there's a few things we can do to make it more reliable and future proof:

- Improve the handling of UID/GID remapping, what we are working on now
  - Make docker behave the same as podman
  - Possibly find an alternative for manually setting a UID/GID different from those of the current linux account. Perhaps we could look into creating a sudo_zwift.sh helper script that could be used to launch zwift as a different user account that starts a proper dbus session, sets up a wayland socket, whatever is needed.
- Update to the latest wine and winetricks version (and keep them up to date)
  - WiP: https://github.com/glennvl/zwift-linux/pull/14
  - Newer wine versions default to EGL instead of GLX
  - We need to figure out how to make EGL work in the container
    - I'm kind of stuck here, I don't know enough about EGL to make this work, I need help!
    - I only tried with nvidia
      - XWayland doesn't work at all
      - The launcher fails to completion with Wayland
      - The game works with Wayland (but we know wine Wayland still has issues)
    - I tried forcing wine to use GLX instead
      - The launcher does not work with GLX at all
      - The game works with GLX and XWayland
- Allow bluetooth access in the container
  - WiP: https://github.com/glennvl/zwift-linux/pull/15
  - Install the bluez driver and add some extra arguments, easy
  - That way we are ready when bluetooth starts working in wine!
- Fix wine dependencies where we can
  - Especially during installation wine spams a lot about missing stuff
  - Look if there's anything we can fix by installing missing dependencies
- Improve documentation
  - We got documentation site on github pages, great!
  - Now we just got to improve the documentation itself
- Release management
  - Updates are done from master, we don't have any official releases
  - I think that ship has sailed, so just leave that as is
  - But we could perhaps add a second unstable branch
    - WiP: https://github.com/glennvl/zwift-linux/pull/16
    - Allow user to opt into unstable branch for updates
    - Push fixes etc immediately to master
    - Push major changes (for example a wine update) to the unstable branch first

That's quite the list, and that's mostly stuff I've already been working on. So there's plenty of hole to fill before we start digging entirely new holes to support other ways of using zwift on linux. :rabbit: :rabbit2: :hole:

Just my opinion of course. :see_no_evil: :sweat_smile:

PR #334, opened 2026-03-22
Upgrade to Wine 11.5 devel with EGL support and fix GPU rendering

pdelagrave:
## Summary

Wine 11 switches from GLX to EGL as the default OpenGL backend. This PR upgrades the image to Wine 11.5 devel and fixes several issues required to get hardware GPU rendering working correctly under Docker.

### Dockerfile
- Bump Wine to `11.5~trixie-1` devel (EGL default since 10.17, Wayland native since 9.9)
- Add `libegl1:i386`, `libgl1:i386`, `libvulkan1:i386` for 32-bit EGL/GL/Vulkan parity with 64-bit
- Bake `10_nvidia.json` into the image to register NVIDIA as a GLVND EGL vendor — the NVIDIA container runtime injects `libEGL_nvidia.so.0` but not the vendor registration file, causing the EGL dispatcher to fall back to Mesa and then software rendering (llvmpipe). Harmless when NVIDIA is absent — GLVND silently skips vendors whose library is not found.

### entrypoint.sh
- **Fix ownership check depth** (`maxdepth 0` → `maxdepth 1`): `mkdir -p $ZWIFT_HOME` runs as root before privilege drop and creates `.wine` owned by root. The previous `maxdepth 0` check on `/home/user` passed (it is owned by user) while missing the root-owned `.wine` subdirectory, causing Wine to refuse to start with "wine: '/home/user/.wine' is not owned by you".
- **Fix gosu supplementary groups**: `gosu user:user` skips `initgroups()` and drops all supplementary groups including the DRI render group. Changed to `gosu user` so all groups from `/etc/group` are loaded.
- **Register DRI render group by GID** before privilege drop using `getent group` to find or create the group by GID, avoiding silent conflict with Mesa's pre-existing `render` group (GID 990).

### update_zwift.sh
- Skip WebView2 install: both the bootstrapper and standalone installer require the `MicrosoftEdgeUpdate` COM service (`{cecddd22-2e72-4832-9606-a9b0e5e344b2}`), which Wine cannot activate (out-of-process COM server for a Windows service). `SilentLaunch` bypasses the launcher UI entirely so WebView2 is not needed at runtime.

### zwift.sh
- Add `--ipc=host` alongside the X11 socket mount so Mesa can attach to X11 shared memory, which is required for DRI2/DRI3 screen initialization.

## Test plan

- [x] First-time install completes without errors
- [x] Zwift launches and renders with hardware GPU (`OpenGL 4.6.0 NVIDIA 570.211.01` on GTX 1070)
- [x] CPU usage normal (~11%) vs software rendering fallback (~430-590%)
- [ ] Test with AMD GPU (radeonsi path)
- [ ] Test with Wayland (`WINE_EXPERIMENTAL_WAYLAND=1`)
- [ ] Test with Podman

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -21,23 +21,22 @@ RUN git clone https://github.com/quietvoid/runfromprocess-rs . \
 
 FROM debian:${DEBIAN_VERSION}-slim AS wine-base
 
-# As at May 2024 Wayland Native works wine 9.9 or later:
-#    WINE_BRANCH="devel"
-# For Specific version fix add WINE_VERSION,
-# make sure to add "=" to the start, comment out for latest
-#    WINE_VERSION="=9.9~bookworm-1"
+# Wine 11: EGL is the default OpenGL backend (replaced GLX since 10.17).
+# Wayland native support included since 9.9. Pinned to 11.5 devel.
+# To update, change WINE_VERSION (prefix with "=", omit for latest devel)
+#    WINE_VERSION="=11.5~${DEBIAN_VERSION}-1"
 ARG DEBIAN_VERSION
 ARG WINE_BRANCH="devel"
-ARG WINE_VERSION="=9.9~${DEBIAN_VERSION}-1"
-ARG WINETRICKS_VERSION=20240105
+ARG WINE_VERSION="=11.5~${DEBIAN_VERSION}-1"
+ARG WINETRICKS_VERSION=20260125
 
 # Install prerequisites
 # - ca-certificates for wget and curl
 # - curl used in zwift authentication script
 # - gamemode for freedesktop screensaver inhibit
 # - gosu for invoking scripts in entrypoint
-# - libegl1 and libgl1 for GL library
-# - libvulkan1 for vulkan loader library
+# - libegl1/libgl1 and i386 variants for GL/EGL library
+# - libvulkan1 and i386 variant for vulkan loader library
 # - procps for pgrep
 # - sudo for normal user installation
 # - wget for downloading winehq key
@@ -51,8 +50,11 @@ RUN dpkg --add-architecture i386 \
         gamemode \
         gosu \
         libegl1 \
+        libegl1:i386 \
         libgl1 \
+        libgl1:i386 \
         libvulkan1 \
+        libvulkan1:i386 \
         procps \
         sudo \
         wget \
@@ -75,11 +77,17 @@ RUN wget -qO /etc/apt/trusted.gpg.d/winehq.asc https://dl.winehq.org/wine-builds
  && chmod +x /usr/local/bin/winetricks
 
 # Create passwordless user and make nvidia libraries discoverable
+# Register NVIDIA as a GLVND EGL vendor so the dispatcher loads libEGL_nvidia.so.0
+# when injected by the NVIDIA container runtime. Harmless if NVIDIA is absent
+# (GLVND silently skips vendors whose library is not found).
 RUN adduser --disabled-password --gecos '' user \
  && adduser user sudo \
  && echo '%SUDO ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers \
  && echo "/usr/local/nvidia/lib" >> /etc/ld.so.conf.d/nvidia.conf \
- && echo "/usr/local/nvidia/lib64" >> /etc/ld.so.conf.d/nvidia.conf
+ && echo "/usr/local/nvidia/lib64" >> /etc/ld.so.conf.d/nvidia.conf \
+ && mkdir -p /usr/share/glvnd/egl_vendor.d \
+ && printf '{"file_format_version":"1.0.0","ICD":{"library_path":"libEGL_nvidia.so.0"}}\n' \
+        > /usr/share/glvnd/egl_vendor.d/10_nvidia.json
 
 # Required for non-glvnd setups
 ENV LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu:/usr/lib/i386-linux-gnu:/usr/local/nvidia/lib:/usr/local/nvidia/lib64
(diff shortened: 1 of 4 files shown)


src/Dockerfile
```diff
@@ -75,11 +77,17 @@ RUN wget -qO /etc/apt/trusted.gpg.d/winehq.asc https://dl.winehq.org/wine-builds
...
+# (GLVND silently skips vendors whose library is not found).
 RUN adduser --disabled-password --gecos '' user \
  && adduser user sudo \
  && echo '%SUDO ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers \
  && echo "/usr/local/nvidia/lib" >> /etc/ld.so.conf.d/nvidia.conf \
- && echo "/usr/local/nvidia/lib64" >> /etc/ld.so.conf.d/nvidia.conf
+ && echo "/usr/local/nvidia/lib64" >> /etc/ld.so.conf.d/nvidia.conf \
+ && mkdir -p /usr/share/glvnd/egl_vendor.d \
+ && printf '{"file_format_version":"1.0.0","ICD":{"library_path":"libEGL_nvidia.so.0"}}\n' \
+        > /usr/share/glvnd/egl_vendor.d/10_nvidia.json
```
glennvl, line 90: Why did you need to add the 10_nvidia.json file? It already exists in the container for me (tried with both docker and podman), I'm guessing this is something the nvidia container toolkit does.

```console
user@glenn-main-pc:/$ ls /usr/share/glvnd/egl_vendor.d/
10_nvidia.json  50_mesa.json
user@glenn-main-pc:/$ cat /usr/share/glvnd/egl_vendor.d/10_nvidia.json 
{
    "file_format_version" : "1.0.0",
    "ICD" : {
        "library_path" : "libEGL_nvidia.so.0"
    }
}
```

src/entrypoint.sh
```diff
@@ -181,7 +181,25 @@ if [[ ${CONTAINER_TOOL} == "docker" ]]; then
...
+            if ! getent group "${dri_gid}" > /dev/null 2>&1; then
+                groupadd -g "${dri_gid}" dri_render 2>/dev/null || true
+            fi
+            dri_group="$(getent group "${dri_gid}" | cut -d: -f1)"
+            if [[ -n ${dri_group} ]]; then
+                usermod -aG "${dri_group}" user 2>/dev/null || true
+                msgbox info "Added user to ${dri_group} group (gid=${dri_gid}) for DRI access"
+            fi
+        fi
+    fi
```
glennvl, line 200: What issue is this trying to fix? Graphics card access already worked?

It does not change anything for me for egl

```text
libEGL warning: egl: failed to create dri2 screen
libEGL warning: egl: failed to create dri2 screen
libEGL warning: egl: failed to create dri2 screen
libEGL warning: egl: failed to create dri2 screen
libEGL warning: failed to open /dev/dri/card1: Permission denied
```

On my machine (fedora 43), the card and render devices belong to a different group:

```console
$ ls -al /dev/dri
total 0
drwxr-xr-x.  3 root root        100 23 mrt 09:23 .
drwxr-xr-x. 21 root root       4960 23 mrt 09:23 ..
drwxr-xr-x.  2 root root        100 23 mrt 09:23 by-path
crw-rw----+  1 root video  226,   1 23 mrt 09:23 card1
crw-rw-rw-.  1 root render 226, 128 23 mrt 09:23 renderD128
```

And the card is /dev/dri/card1, so we shouldn't assume it's card0?

src/update_zwift.sh
```diff
@@ -137,10 +137,9 @@ install_zwift() {
     msgbox info "Installing prerequisites using winetricks"
     winetricks -q dotnet20 dotnet48 d3dcompiler_47 || return 1
 
-    # download and install webview 2
-    msgbox info "Downloading and installing webview2"
-    wget -O webview2-setup.exe https://go.microsoft.com/fwlink/p/?LinkId=2124703 || return 1
-    wine webview2-setup.exe /silent /install || return 1
+    # WebView2 install skipped: both bootstrapper and standalone require the
+    # MicrosoftEdgeUpdate COM service which Wine cannot activate. SilentLaunch
+    # bypasses the launcher UI so WebView2 is not needed at runtime.
```
glennvl, line 142: This appears to be a specific issue with wine 11.5. It still worked with wine 11.3. But if we don't need webview2, we might indeed as well remove it.

glennvl (commented):
I tried with both docker and podman and I still have the same issues as I have in my own fork https://github.com/glennvl/zwift-linux/pull/14 if I don't add `--ipc=host`.

eglgears_x11 still says `MESA: error: Failed to attach to x11 shm` even with `--ipc=host`.

### docker

- With docker zwift launches with software rendering if I add `--ipc=host` to the zwift and build scripts.
- Adding the container user to the render group does not make a difference.

```text
libEGL warning: egl: failed to create dri2 screen
libEGL warning: failed to open /dev/dri/card1: Permission denied
```

### podman

With podman zwift launches with software rendering if I add `--ipc=host` to the zwift and build scripts.

```text
libEGL warning: egl: failed to create dri2 screen
```

src/Dockerfile
```diff
@@ -51,8 +50,11 @@ RUN dpkg --add-architecture i386 \
         gamemode \
         gosu \
         libegl1 \
+        libegl1:i386 \
         libgl1 \
+        libgl1:i386 \
         libvulkan1 \
+        libvulkan1:i386 \
```
glennvl, line 57: Why did you add the 32-bit libraries? They're not needed?

src/entrypoint.sh
glennvl: With docker, if I add the user to the group from /dev/dri/card1 and launch with gosu user instead of gosu user:user, the permission denied warning goes away. But the other warning remains and it still uses software rendering instead of hardware rendering.

PR #341, opened 2026-04-13
Fix zwift install error with docker

glennvl:
## Summary

Checking the ownership of only the mount point itself appears to be not enough.

Fixes zwift install error when using docker:

```text
wine: '/home/user/.wine' is not owned by you.
```

This is probably also the reason why the build from scratch workflow failed.

I remember that we discussed that this change was needed a while ago. And I thought it was actually already implemented. But I can't seem to find it in the git history, issues, discussions or pull requests.

### DIFF
--- a/src/entrypoint.sh
+++ b/src/entrypoint.sh
@@ -142,7 +142,7 @@ if [[ ${CONTAINER_TOOL} == "docker" ]]; then
         # This avoids a costly recursive find on every normal startup
         local target="${1:?}"
         local result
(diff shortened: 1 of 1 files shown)


netbrain:
Hm weird..

glennvl:
> Hm weird..

Found where I read it :slightly_smiling_face:.

@pdelagrave discovered this issue and mentioned it in #334.

> ### entrypoint.sh
> - **Fix ownership check depth** (`maxdepth 0` → `maxdepth 1`): `mkdir -p $ZWIFT_HOME` runs as root before privilege drop and creates `.wine` owned by root. The previous `maxdepth 0` check on `/home/user` passed (it is owned by user) while missing the root-owned `.wine` subdirectory, causing Wine to refuse to start with "wine: '/home/user/.wine' is not owned by you".

PR #342, opened 2026-04-15
Don't start an interactive shell when ZWIFT_FG is set

kschat:
This is a regression that occurred a while ago, but I just noticed it because I was lazy and haven't updated `zwift.sh` in months. See #213 for the use-case (I also updated the docs to include my use-case :smile:).


glennvl (approved):

kschat:
Thanks for the quick review and merge, @glennvl!

PR #343, opened 2026-04-16
ZWIFT_FG and interactive shell

glennvl:
## Summary

See #342 for context.

When running zwift in the foreground (`ZWIF_FG=1`):

- Adding `-it` caused a regression as described in #342, so we removed it again.
- Without `-it` it's not possible to use for example `ctrl+c` to terminate zwift.

Can we do both? This PR only adds the `-it` flag when `ZWIFT_FG=1` if the zwift.sh script is running in an interactive terminal.

## Implementation

- We already check whether the zwift script is running in an interactive terminal using `[[ -t 1 ]]`.
- Based on that we enable or disable fancy messages.
- Set a flag `INTERACTIVE_TERMINAL=1` and use it to enable `-it` if `ZWIFT_FG=1`.

## Testing

@kschat Could you try if this works for your usecase?


netbrain:
Seems reasonable

kschat:
Sorry for the delay. I just tested this change and launching from Steam still works.

Issue #285, opened 2026-02-09
Extremely low framerate after latest update (WINE_EXPERIMENTAL_WAYLAND=1)


trahflow:
After one of the latest updates (both, the script and the image were updated), the framerate droppend to extremely low values (1-2 FPS).
No configuration changes were done in between.
ENV:
```bash
VGA_DEVICE_FLAG="--runtime=nvidia --gpus=all"
DEBUG=1
CONTAINER_TOOL=docker
ZWIFT_FG=1
WINE_EXPERIMENTAL_WAYLAND=1
```
System: Arch Linux x86_64 on NVIDIA RTX 2060
```bash
$ cat /proc/driver/nvidia/version
NVRM version: NVIDIA UNIX Open Kernel Module for x86_64  590.48.01  Release Build  (root@)  
GCC version:  gcc version 15.2.1 20260103 (GCC) 
```

Not setting `WINE_EXPERIMENTAL_WAYLAND=1` does bring the performance back to normal level, but setting that flag did work on previous versions.
Posting this to provide a data point, and maybe it's helpful to others.

glennvl:
In the container where I got it to work, I get about half the fps with wayland (~70fps) vs xwayland (~130fps). In another container it just crashed with an opengl error. Wayland also doesn't properly apply my host system style. So I think it's safe to say that the experimental is there for a reason.

glennvl:
Have you tried updating to the latest version that we pushed yesterday? It's using a new version of debian base image. Did anything change for you with regards to performance? Is wayland still working or did it break? Is your experience similar to #287?

trahflow:
> Have you tried updating to the latest version that we pushed yesterday? 

Just did so.
> Did anything change for you with regards to performance?

Still the same as reported. Performance with `WINE_EXPERIMENTAL_WAYLAND=1` is ~2 FPS
> Is wayland still working or did it break?

It's working, but dead slow.
> Is your experience similar to [#287](https://github.com/netbrain/zwift/issues/287)?

No crashes. No wine error as shown in the screenshot there. Just very slow.

msouthcott:
I have the same issue but I don't have NVIDIA.
Graphics:
  Device-1: Intel WhiskeyLake-U GT2 [UHD Graphics 620] vendor: Lenovo
    driver: i915 v: kernel arch: Gen-9.5 process: Intel 14nm built: 2016-20
[...]

PR #344, opened 2026-04-17
Security hardening: feature-flag privileged mode, strip sudo, extend HTTP/timeout

philipbalinov:
## Summary

Security audit identified several issues in the container setup. This PR addresses the most critical ones, all verified working with a successful local build and game launch.

- **Remove `--privileged` default** on non-SELinux systems — GPU access is already provided via `--device=/dev/dri` flags; `--privileged` granted unnecessary full host device access and container escape paths
- **Remove passwordless `sudo`** (`NOPASSWD:ALL`) and user sudo group membership — the entrypoint already uses `gosu` for privilege drops; sudo was unused and allowed trivial root escalation from any code execution inside the container
- **Verify winetricks SHA256** at build time — adds `WINETRICKS_SHA256` ARG and `sha256sum --check` step so a compromised or MITM'd winetricks download fails the build rather than silently executing as root
- **Add `mesa-vulkan-drivers`** — `VK_KHR_surface`/`VK_KHR_win32_surface` were missing inside the container, causing ZwiftLauncher's UI to fail to initialise; this also fixes the `wglSetPixelFormatWINE` errors
- **Switch version-check URL from HTTP to HTTPS** (`update_zwift.sh`) — prevents on-path attackers from manipulating the version XML to trigger malicious updates
- **Download `install.sh` to a temp file** before executing via `pkexec` instead of piping `curl` output directly into `bash -c` — the script is now on disk and inspectable before execution
- **Increase ZwiftLauncher update timeout** from 5 min to 60 min — the full game asset download is ~4 GB and consistently exceeded the previous limit

## Test plan

- [x] `podman build` completes successfully with winetricks SHA256 verified
- [x] ZwiftLauncher downloads and installs game assets to v1.111.0
- [x] `podman commit` succeeds and image is tagged
- [x] `zwift.sh` launches the game successfully without `--privileged`

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -39,7 +39,6 @@ ARG WINETRICKS_VERSION=20240105
 # - libegl1 and libgl1 for GL library
 # - libvulkan1 for vulkan loader library
 # - procps for pgrep
-# - sudo for normal user installation
 # - wget for downloading winehq key
 # - winbind for ntml_auth required by zwift/wine
 # - xdg-utils seems to be a dependency of wayland
@@ -54,7 +53,6 @@ RUN dpkg --add-architecture i386 \
         libgl1 \
         libvulkan1 \
         procps \
-        sudo \
         wget \
         winbind \
         xdg-utils \
@@ -74,10 +72,8 @@ RUN wget -qO /etc/apt/trusted.gpg.d/winehq.asc https://dl.winehq.org/wine-builds
  && wget -qO /usr/local/bin/winetricks https://raw.githubusercontent.com/Winetricks/winetricks/${WINETRICKS_VERSION}/src/winetricks \
  && chmod +x /usr/local/bin/winetricks
 
-# Create passwordless user and make nvidia libraries discoverable
+# Create user and make nvidia libraries discoverable
 RUN adduser --disabled-password --gecos '' user \
- && adduser user sudo \
- && echo '%SUDO ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers \
  && echo "/usr/local/nvidia/lib" >> /etc/ld.so.conf.d/nvidia.conf \
  && echo "/usr/local/nvidia/lib64" >> /etc/ld.so.conf.d/nvidia.conf
 
--- a/src/update_zwift.sh
+++ b/src/update_zwift.sh
@@ -67,7 +67,7 @@ get_current_version() {
 }
 
 get_latest_version() {
-    wget --no-cache --quiet -O - http://cdn.zwift.com/gameassets/Zwift_Updates_Root/Zwift_ver_cur.xml \
+    wget --no-cache --quiet -O - https://cdn.zwift.com/gameassets/Zwift_Updates_Root/Zwift_ver_cur.xml \
         | grep -oP 'sversion="\K.*?(?=")' | cut -f 1 -d ' ' \
         || return 1
 }
@@ -96,7 +96,7 @@ update_zwift_using_launcher() {
     msgbox ok "Zwift launcher started using wine"
 
     local counter=1
-    local max_iterations=60 # 60 * 5s = 5 minutes max
+    local max_iterations=720 # 720 * 5s = 60 minutes max
     # also stop if launcher exits before update finishes, so we don't hang forever
     while [[ ${zwift_current_version} != "${zwift_latest_version}" ]] && pgrep -f ZwiftLauncher.exe > /dev/null 2>&1; do
         if [[ ${counter} -gt ${max_iterations} ]]; then
--- a/src/zwift.sh
+++ b/src/zwift.sh
@@ -239,16 +239,21 @@ check_script_up_to_date() {
 }
 
 upgrade_script() {
-    local install_script
+    local install_script_file
+    install_script_file="$(mktemp -q /tmp/zwift-install.XXXXXXXXXX.sh)" || {
+        msgbox error "Failed to create temporary file for install script"
+        return 1
+    }
+    trap 'rm -f -- "${install_script_file}"' RETURN
 
     msgbox info "Downloading latest install script"
-    if ! install_script="$(curl -fsSL https://raw.githubusercontent.com/netbrain/zwift/master/bin/install.sh)"; then
+    if ! curl -fsSL https://raw.githubusercontent.com/netbrain/zwift/master/bin/install.sh -o "${install_script_file}"; then
         msgbox error "Failed to download install script"
         return 1
     fi
 
-    msgbox info "Running install script"
-    if ! pkexec env PATH="${PATH}" bash -c "${install_script}" -- --script-version="${SCRIPT_VERSION}"; then
+    msgbox info "Running install script (inspect ${install_script_file} before proceeding)"
+    if ! pkexec env PATH="${PATH}" bash "${install_script_file}" -- --script-version="${SCRIPT_VERSION}"; then
         msgbox error "Install script failed"
         return 1
     fi
@@ -486,8 +491,7 @@ elif is_selinux_active; then
     msgbox info "SELinux is active, using secure container flags"
     container_args+=(--security-opt label=type:container_runtime_t)
 else
-    msgbox warning "Not using SELinux, running container in privileged mode to be able to access the GPU"
-    container_args+=(--privileged --security-opt label=disable)
+    msgbox info "Running container without privileged mode; GPU access is provided via --device flags below"
 fi
 
 # Append extra arguments provided by user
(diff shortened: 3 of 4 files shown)


src/Dockerfile
```diff
@@ -30,6 +30,9 @@ ARG DEBIAN_VERSION
 ARG WINE_BRANCH="devel"
 ARG WINE_VERSION="=9.9~${DEBIAN_VERSION}-1"
 ARG WINETRICKS_VERSION=20240105
+# SHA256 of the winetricks script for the version above.
+# To update: curl -fsSL https://raw.githubusercontent.com/Winetricks/winetricks/<VERSION>/src/winetricks | sha256sum
+ARG WINETRICKS_SHA256=8f28985b06e9e04b5268e2aef3429b2a0f6abd4fd12c4a30472dfe66355128fa
```
glennvl: Discussed in #330.

Conclusion:

* We will not be adding a sha for winetricks.
* We trust the combination of https and pinning the version.

src/update_zwift.sh
```diff
@@ -67,7 +67,7 @@ get_current_version() {
 }
 
 get_latest_version() {
-    wget --no-cache --quiet -O - http://cdn.zwift.com/gameassets/Zwift_Updates_Root/Zwift_ver_cur.xml \
+    wget --no-cache --quiet -O - https://cdn.zwift.com/gameassets/Zwift_Updates_Root/Zwift_ver_cur.xml \
```
glennvl, line 70: Good catch. This should indeed be https.

src/update_zwift.sh
```diff
@@ -96,7 +96,7 @@ update_zwift_using_launcher() {
     msgbox ok "Zwift launcher started using wine"
 
     local counter=1
-    local max_iterations=60 # 60 * 5s = 5 minutes max
+    local max_iterations=720 # 720 * 5s = 60 minutes max
```
glennvl, line 99: Did you have an issue with the timeout not being long enough? Waiting for an hour seems excessive. In the workflow this takes about 30 of the 60 iterations that were already allowed, so that appears to be a safe margin already.

src/zwift.sh
```diff
@@ -239,16 +239,21 @@ check_script_up_to_date() {
...
-    if ! install_script="$(curl -fsSL https://raw.githubusercontent.com/netbrain/zwift/master/bin/install.sh)"; then
+    if ! curl -fsSL https://raw.githubusercontent.com/netbrain/zwift/master/bin/install.sh -o "${install_script_file}"; then
         msgbox error "Failed to download install script"
         return 1
     fi
 
-    msgbox info "Running install script"
-    if ! pkexec env PATH="${PATH}" bash -c "${install_script}" -- --script-version="${SCRIPT_VERSION}"; then
+    msgbox info "Running install script (inspect ${install_script_file} before proceeding)"
+    if ! pkexec env PATH="${PATH}" bash "${install_script_file}" -- --script-version="${SCRIPT_VERSION}"; then
```
glennvl, line 257: Why did you change this?

src/zwift.sh
```diff
@@ -486,8 +491,8 @@ elif is_selinux_active; then
     msgbox info "SELinux is active, using secure container flags"
     container_args+=(--security-opt label=type:container_runtime_t)
 else
-    msgbox warning "Not using SELinux, running container in privileged mode to be able to access the GPU"
-    container_args+=(--privileged --security-opt label=disable)
+    msgbox info "Running container without privileged mode; GPU access is provided via --device flags below"
+    container_args+=(--security-opt no-new-privileges:true)
```
glennvl: Enabling privileged mode here was added in #336 to fix #285. How did you test this?

(In my personal opinion the end-user should set PRIVILEGED_MODE=1 himself in this situation, but that's a different discussion.)

src/Dockerfile
```diff
@@ -71,13 +73,13 @@ RUN wget -qO /etc/apt/trusted.gpg.d/winehq.asc https://dl.winehq.org/wine-builds
...
+ && wget -qO /tmp/winetricks https://raw.githubusercontent.com/Winetricks/winetricks/${WINETRICKS_VERSION}/src/winetricks \
+ && echo "${WINETRICKS_SHA256}  /tmp/winetricks" | sha256sum --check \
+ && mv /tmp/winetricks /usr/local/bin/winetricks \
  && chmod +x /usr/local/bin/winetricks
 
-# Create passwordless user and make nvidia libraries discoverable
+# Create user and make nvidia libraries discoverable
 RUN adduser --disabled-password --gecos '' user \
- && adduser user sudo \
- && echo '%SUDO ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers \
```
glennvl, line 80: Probably safe to remove.

src/Dockerfile
```diff
@@ -53,8 +55,8 @@ RUN dpkg --add-architecture i386 \
         libegl1 \
         libgl1 \
         libvulkan1 \
+        mesa-vulkan-drivers \
```
glennvl: - You claim that this fixes an issue with the zwift launcher ui. Can you show a before and after screenshot?
- Can you show the `wglSetPixelFormat` errors you were getting that are now solved?

glennvl (changes requested):
You used a large language model to modify source code. I'm going to need evidence that the changes are useful and fix rather than break stuff.

src/update_zwift.sh
philipbalinov: apparently - Zwift should really start using a CDN. I'm shocked any update downloads in less than 10 minutes, I've missed rides because Zwift was updating, and I'm on a fast internet connection.

src/zwift.sh
philipbalinov: - line 250 tells curl to write to a file instead of storing curl STDOUT in a variable
- line 255 is an optional hint for the user for more transparency what is being executed
- line 256 removes the `-c` since we're calling a script file, not inlining commands

src/Dockerfile
philipbalinov: that was part of another PR that got mixed up, removed

src/Dockerfile
philipbalinov: reverted the change

src/update_zwift.sh
philipbalinov: well apparently they do cloudfront, but it's still painfully slow for me; looking into why, perhaps it's a wine thing

src/zwift.sh
philipbalinov: interesting, i run on Debian 13 without SELinux and have never had a problem running unprivileged - but then my user is part of the `video` group, and i use podman - maybe that has an effect
Regardless, since I have no way of turning the privileged mode off right now, i propose a new flag
`ZWIFT_NO_PRIVILEGED` that will be off by default; 

alternatively we can refactor `PRIVILEGED_MODE` so that it covers all uses cases and gives non-SELinux a way of dropping privileges

src/Dockerfile
netbrain: Been mostly used for debugging, ease of use..

src/update_zwift.sh
glennvl: > I've missed rides because Zwift was updating

The update/install code only runs in a github workflow, it does not run on your pc.

src/zwift.sh
glennvl: But why?

glennvl:
> - **Remove passwordless `sudo`** (`NOPASSWD:ALL`) and user sudo group membership — the entrypoint already uses `gosu` for privilege drops; sudo was unused and allowed trivial root escalation from any code execution inside the container
> - **Verify winetricks SHA256** at build time — adds `WINETRICKS_SHA256` ARG and `sha256sum --check` step so a compromised or MITM'd winetricks download fails the build rather than silently executing as root
> - **Switch version-check URL from HTTP to HTTPS** (`update_zwift.sh`) — prevents on-path attackers from manipulating the version XML to trigger malicious updates

Implemented in #347

> - **Remove `--privileged` default** on non-SELinux systems — GPU access is already provided via `--device=/dev/dri` flags; `--privileged` granted unnecessary full host device access and container escape paths

Not changing this. I agree, but this was specifically added in #336 to fix #285.

> - **Add `mesa-vulkan-drivers`** — `VK_KHR_surface`/`VK_KHR_win32_surface` were missing inside the container, causing ZwiftLauncher's UI to fail to initialise; this also fixes the `wglSetPixelFormatWINE` errors

Not changing this. Adding mesa-vulkan-drivers does not fix wglSetPixelFormatWINE,.

> - **Download `install.sh` to a temp file** before executing via `pkexec` instead of piping `curl` output directly into `bash -c` — the script is now on disk and inspectable before execution

Not changing this. It's already possible to inspect the script contents by setting DEBUG=1.

> - **Increase ZwiftLauncher update timeout** from 5 min to 60 min — the full game asset download is ~4 GB and consistently exceeded the previous limit

Not changing this. Misunderstanding of the timeout, it is used only in the github action and long enough.

PR #347, opened 2026-04-20
Feature/wine-stability

glennvl:
## Summary

Make some changes with the goal of improving stability.

- Update winetricks and dependencies installed with winetricks
  - Winetricks 20260125, webview2 in the zwift launcher works!
  - Remove dotnot20, not needed anymore
  - Add corefonts, fixes wine warnings about missing fonts
- Wine
  - ~Install wine mono and gecko (instead of launching wine with `WINEDLLOVERRIDES="mscoree,mshtml="`)~
  - ~This fixes .net framework compilation errors~
- Dockerfile
  - Update rust base image
  - [Use ADD instruction instead of RUN wget](https://docs.docker.com/build/building/best-practices/#add-or-copy)
  - Use checksums for downloads (this has come up a few times, but it's actually easy with ADD, so no real reason anymore to avoid doing this)
  - Use ADD to clone the runfromprocess git repo at the desired commit
  - Add `DEBIAN_FRONTEND="noninteractive"` to apt-get commands to prevent dialog creation errors
- Refactor run_zwift.sh
  - Launch wineserver first, so it doesn't automatically start first and then gets killed and relaunched with gamemode
  - Use the windows task manager with wine to manage wine processes, keeping windows stuff in windows
  - Wait for wine processes where possible
  - Kill the zwift launcher as soon as possible, this essentially kills the launcher immediately, so it won't have the time to start update processes etc, effectively bypassing it entirely
- Add update-image.sh script
  - Similar to the workflow, it assumes the image exists and updates the scripts in it
  - Makes it quicker to work on the container scripts. Only run build-image.sh once (slow), then use update-image.sh when making additional changes to the scripts (quick).
- Minor changes
  - Optimize the msgbox function, reducing the number of lines of code
  - Fix the zwift version file url (use https instead of http)
  - Don't add the container user to the sudo group. It doesn't work with podman anyway because we use `--userns=keep`. Easy to get root access with `$CONTAINER_TOOL exec -it -u root zwift-$USER bash`.

@philipbalinov this includes some of the stuff you proposed.

## Discussion

I've been experimenting with gamemode to see what it is actually doing and how well it is working. As it stands now, the wineserver is started in gamemode. And gamemode is indeed active as long as the wineserver is running. This will prevent sleep, screensaver, ..., but it won't do anything to optimize performance.

If we want to use gamemode to optimize performance, we have to launch zwift itself with gamemode. I've added the missing 32-bit gamemode libraries that appear to be required to do so. But it's currently not possible because zwift is started with runfromprocess. If we start runfromprocess in gamemode, gamemode will activate, optimize runfromprocess, optimize zwift, and then when runfromprocess stops (zwift has launched) immediately remove the optimizations and turn off again.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -70,16 +70,8 @@ msgbox() {
     local type="${1:?}" # Type: info, ok, warning, error, question, debug
     local msg="${2:?}"  # Message: the message to display
 
-    make_timestamp() {
-        if [[ ${VERBOSITY} -ge 2 ]]; then
-            printf '%(%T)T|' -1
-        else
-            printf ''
-        fi
-    }
-
-    local timestamp
-    timestamp="$(make_timestamp)"
+    local timestamp=""
+    [[ ${VERBOSITY} -ge 2 ]] && printf -v timestamp '%(%T)T|' -1
 
     case ${type} in
         info) [[ ${VERBOSITY} -ge 1 ]] && echo -e "${COLOR_BLUE}[${timestamp}*] ${msg}${RESET_STYLE}" ;;
--- a/docs/contributing/development.md
+++ b/docs/contributing/development.md
@@ -128,6 +128,9 @@ podman build -t zwift:dev .
 
 # Or use the build script
 ./build-image.sh
+
+# Or only update the scripts if you already built the imgage
+./update-image.sh
 ```
 
 ## Testing Changes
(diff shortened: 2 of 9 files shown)


netbrain:
LGTM, cant find anything to fault.

netbrain (approved):

PR #348, opened 2026-04-27
Factorize duplicated setup code into lib.sh

avandecreme:
Extract duplicated DEBUG, color variables, and msgbox function from entrypoint.sh, run_zwift.sh, update_zwift.sh and zwift-auth.sh into a shared lib.sh.

This is extracted from #345 to ease review

### DIFF
--- a/src/.shellcheckrc
+++ b/src/.shellcheckrc
@@ -1,0 +1,1 @@
+source-path=SCRIPTDIR
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -98,6 +98,7 @@ LABEL org.opencontainers.image.title="netbrain/zwift"
 LABEL org.opencontainers.image.description="Easily zwift on linux"
 LABEL org.opencontainers.image.url="https://github.com/netbrain/zwift"
 
+COPY --chmod=644 lib.sh /bin/lib.sh
 COPY --chmod=755 entrypoint.sh /bin/entrypoint
 COPY --chmod=755 update_zwift.sh /bin/update_zwift.sh
 COPY --chmod=755 run_zwift.sh /bin/run_zwift.sh
--- a/src/entrypoint.sh
+++ b/src/entrypoint.sh
@@ -1,27 +1,9 @@
 #!/usr/bin/env bash
 set -uo pipefail
 
-readonly DEBUG="${DEBUG:-0}"
-if [[ ${DEBUG} -eq 1 ]]; then set -x; fi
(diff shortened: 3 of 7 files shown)


glennvl:
I've been thinking about how to fix the duplicated code too.

There's two aspects to it.

1. The duplicated code shared by install.sh and zwift.sh
2. The duplicated code shared by the scripts that live inside the container image

As for 1. extracting the duplicated code in a library would introduce additional complexity. It's much more convenient to keep the install.sh and zwift.sh scripts standalone.

The duplicated code in 2. is something we can fix. But if we do so, I'd rather see it done properly. I would probably create a couple of libraries and move all the container scripts to /opt so they don't pollute /bin too much and are all together in a distinct directory. A possible structure could be (just an initial idea, I don't know if those four libs make sense):

```text
.
└── /opt
    ├── /netbrain
    │   └── /zwift
    │       ├── /lib
    │       │   ├── logging.sh     : DEBUG, set -x, VERBOSITY, colors, msgbox, ...
    │       │   ├── utils.sh       : wait_until, wait_until_process_started, ...
    │       │   ├── wine.sh        : wait_until_wine_task_started, wait_until_wine_task_stopped, ...
    │       │   └── zwift.sh       : WINE_USER_HOME, ZWIFT_HOME, ZWIFT_DOCS, ZWIFT_PREFS, ...
    │       ├── authenticate.sh
    │       ├── entrypoint.sh
    │       ├── install.sh
    │       └── launch.sh
    └── /quietvoid
        └── /rfp-rs
            └── /run-from.exe
```

(Note: Don't forget to update the github workflows and documentation)

At that point I would probably move a few files around in the project as well, to better distinguish between helper scripts, scripts that are meant for the host, and stuff that lives in the container image.

- Create a top-level scripts directory for helper scripts (build-image.sh and update-image.sh from #347).
- Container image related stuff stays in src: Dockerfile, pulse-client.conf, scripts. We could call the libraries lib-XXX.sh so we don't need an additional src/lib directory.
- What do we do with the zwift.sh script? Figure out a way to distinguish between container and host stuff in the src directory? Or move it to the bin directory with install.sh?

---

@netbrain What do you think?

@avandecreme I we go through with this, I can help you if you want to give it a try, or I can do it.

avandecreme:
Oops, completely missed that it wouldn't work outside of nix for zwift.sh.

You can go ahead with the refactoring you have in mind. I will remove that change from #345 and close this.

Issue #356, opened 2026-05-13
support for non-interactive zwift.sh / Steam integration


BreiteSeite:
## Describe the bug

I run zwift on a HTPC that is connected to my TV and runs steam big picture. Hence i start zwift.sh via steam (thanks to [steamgriddb](https://www.steamgriddb.com/search/grids?term=Zwift) it looks more or less like a native steam game).

However, there are two "issues" to make this setup work smoothly as far as i can see.

## Interactive Input prompts
zwift.sh prompts for inputs. It asks if it should update zwift.sh and/or the container image (y/N questions). These prompts seem to make it impossible to launch it via steam, as it waits for input and steam kills the process at some point.

It would be nice to have two ENV variables where i can specify that i always want to update both (zwift.sh and container image). (Ideally with some small window showing update progress so launching in steam and waiting it to update does not have me to silently wait without knowing what's going on. Maybe firing off (`notify-send` could be a quick-win - however, i don't know if it is shown in steam big picture mode).

Alternatively: Enable an "update only and exit" launch mode, so i can run a systemd.timer that does it regularly without launching zwift. Then i could use the existing DONT_CHECK/DONT_PULL environment variables to always launch it without updating and prompting

## Permission escalation
When updating zwift.sh, it prompts me for a root password. It would be nice to have it possible to have zwift.sh installed for the current user only, so no extra permissions are necessary - as this is difficult to enter when using a controller. Or is there a way to automatically have keychain authorize this?

Thank you for your help!

And big thanks for your great work containerizing zwift. I gave up installing zwift manually via lutris and wine after a couple of hours, so frustrating! This project is a godsend for linux-based zwifters.

glennvl:
> However, there are two "issues" to make this setup work smoothly as far as i can see.
>
> 1. zwift.sh prompts for inputs. It asks if it should update zwift.sh and/or the container image (y/N questions).
> 2. When updating zwift.sh, it prompts me for a root password. It would be nice to have it possible to have zwift.sh installed for the current user only, so no extra permissions are necessary

As for the second point. The zwift install scripts allows installing for the current user only. But the automatic update procedure always launches the installer with the root prompt. This is an issue we are aware of. The zwift script should be modified to detect whether it's installed as root or current user instead of always updating as root.

We could add a parameter (evironment variable or script argument) to auto confirm questions, but I think we shouldn't until the second point is fixed.

[...]

PR #357, opened 2026-05-14
Fix updating when installed as user

glennvl:
## Summary

Related to #356.

Fix automatic zwift script update forcing system wide install

- If netbrain/zwift is installed for the local user, don't use pkexec when invoking the install script in the update procedure.
- Add an uninstall option to the install script. This makes it easier for users to switch from a system wide install to a user local install if they want to.
- Recommend a user local installation instead of a system wide installation in the documentation.

Fix unable to update zwift script in non-interactive environment

- In a non-interactive environment, the question to confirm the zwift script update will timeout and default to false. The zwift script will therefore never be updated, while the container image will.
- Instead, assume yes in a non-interactive environment. The user can set DONT_CHECK/DONT_PULL to disable updates.

## Details

Uninstalling netbrain/zwift will:

- Remove the scripts and desktop entry
- Keep the Zwift container image
- Keep all configuration (volume and config files)

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -69,6 +69,7 @@ readonly ZWIFT_LOGO="https://raw.githubusercontent.com/netbrain/zwift/master/bin
 readonly ZWIFT_DESKTOP_ENTRY="https://raw.githubusercontent.com/netbrain/zwift/master/bin/Zwift.desktop"
 
 if [[ -t 1 ]]; then
+    readonly INTERACTIVE_TERMINAL=1
     readonly COLOR_WHITE="\033[0;37m"
     readonly COLOR_RED="\033[0;31m"
     readonly COLOR_GREEN="\033[0;32m"
@@ -78,6 +79,7 @@ if [[ -t 1 ]]; then
     readonly STYLE_UNDERLINE="\033[4m"
     readonly RESET_STYLE="\033[0m"
 else
+    readonly INTERACTIVE_TERMINAL=0
     readonly COLOR_WHITE=""
     readonly COLOR_RED=""
     readonly COLOR_GREEN=""
@@ -274,6 +276,11 @@ install_netbrain_zwift() {
 
 echo -e "${COLOR_YELLOW}[!] ${STYLE_BOLD}Easily Zwift on linux!${RESET_STYLE}"
 echo -e "${COLOR_YELLOW}[!] ${STYLE_UNDERLINE}https://github.com/netbrain/zwift${RESET_STYLE}"
+
+if [[ ${INTERACTIVE_TERMINAL} -eq 0 ]] && [[ ${auto_confirm} -eq 0 ]]; then
+    msgbox warning "Detected non-interactive environment, enabling auto-confirm"
+    auto_confirm=1
+fi
 
 if [[ ${uninstall} -eq 1 ]]; then
     uninstall_netbrain_zwift
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -5,15 +5,29 @@ readonly DEBUG="${DEBUG:-0}"
 if [[ ${DEBUG} -eq 1 ]]; then set -x; fi
 
 print_usage() {
-    echo "Usage: install [ -v | --script-version COMMIT_HASH ]"
-    echo "               [ -y | --auto-confirm ]"
-    echo "               [ -h | --help ]"
+    echo "Install or uninstall netbrain/zwift"
+    echo "Invoking this script with sudo will perform a system wide install"
+    echo "Invoking this script without sudo will perform a user local install"
+    echo ""
+    echo "Usage:"
+    echo "    ${0} [ -v | --script-version COMMIT_HASH ]"
+    echo "    ${0//?/ } [ -y | --auto-confirm ]"
+    echo "    ${0//?/ } [ -u | --uninstall ] "
+    echo "    ${0//?/ } [ -h | --help ]"
+    echo ""
+    echo "Options:"
+    echo "    --script-version COMMIT_HASH    Install a specific netbrain/zwift version instead of master"
+    echo "    --auto-confirm                  Automatically confirm installation"
+    echo "    --uninstall                     Uninstall netbrain/zwift"
+    echo "    --help                          Print usage"
     exit 2
 }
 
+script_args=("${@}")
 script_version="master"
 auto_confirm=0
-if ! options="$(getopt -n install -o "v:yh" -l "script-version:,auto-confirm,help" -- "${@}")"; then
+uninstall=0
+if ! options="$(getopt -n install -o "v:yuh" -l "script-version:,auto-confirm,uninstall,help" -- "${@}")"; then
     print_usage
 fi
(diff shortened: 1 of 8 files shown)


bin/install.sh
```diff
@@ -94,105 +118,172 @@ msgbox() {
...
+        fi
+    }
+
+    remove_user_install() {
+        if [[ -f "${USER_BIN_DIR}/zwift" ]]; then
+            if msgbox question "Found user install of netbrain/zwift, remove it?"; then
+                msgbox ok "Removing user install"
+                remove_file "${USER_BIN_DIR}/zwift"
+                remove_file "${USER_SHARE_DIR}/icons/hicolor/scalable/apps/zwift.svg"
+                remove_file "${USER_SHARE_DIR}/applications/Zwift.desktop}"
```
netbrain: **Bug — trailing `}` on these paths (here and line 158):**

```bash
remove_file "${USER_SHARE_DIR}/applications/Zwift.desktop}"
remove_file "${SYSTEM_SHARE_DIR}/applications/Zwift.desktop}"
```

`rm -f` won't complain about the bad path, so the desktop entry is never removed on uninstall. Drop the trailing `}`.

netbrain:
Suggestion for the Nix install path: `flake.nix` already exposes a `dontCheck` option on the NixOS module (defaults to `false`). Flipping the default to `true` would make Nix-flake users skip the self-update check by default — which is what they want, since the script lives under `/nix/store/...` and the in-script updater would either fail (read-only store) or silently write a duplicate `~/.local/bin/zwift` that shadows the Nix-managed one on `PATH`. Users who want the old behavior can still set `programs.zwift.dontCheck = false;`.

Probably worth a separate PR rather than scope-creeping this one — flagging here for visibility.

bin/install.sh
glennvl: And now I feel like an idiot for missing that. Thanks for the careful review! Fixed in b193b17731c51e5750b0f3e14211f392bb886f5a.

glennvl:
> Suggestion for the Nix install path: `flake.nix` already exposes a `dontCheck` option on the NixOS module (defaults to `false`). Flipping the default to `true` would make Nix-flake users skip the self-update check by default — which is what they want, since the script lives under `/nix/store/...` and the in-script updater would either fail (read-only store) or silently write a duplicate `~/.local/bin/zwift` that shadows the Nix-managed one on `PATH`. Users who want the old behavior can still set `programs.zwift.dontCheck = false;`.
> 
> Probably worth a separate PR rather than scope-creeping this one — flagging here for visibility.

Hmm, sounds like we need to figure out a way to correctly update the zwift script when using nix then. Agree that this should be done in a separate PR. Maybe make an issue where we can think about how to properly solve this first?

bin/install.sh
glennvl: I also changed the remove_file function in 27f4fc9985c25bed13aefd758e53b9bfa70f489b:

```bash
  remove_file() {
      local file="${1:?}"

      msgbox info "  Removing ${file}"

      if ! rm -- "${file}" > /dev/null 2>&1; then
          msgbox warning "  Failed to remove ${file}"
      fi
  }
```

glennvl:
@netbrain I added two more commits with some documentation changes d66c4d1c6c2671ee266c9413b14d23d28864dc0e and 4f8585a393e4d6e6925253b37724da3c49d90622. But this PR has been sitting here open for a while now, time to merge?

netbrain:
Sorry about that. Been super busy as of late 🫩...

I'll take a look at it later. On the move right now..

src/zwift.sh
```diff
@@ -57,6 +60,7 @@ msgbox() {
         warning) echo -e "${COLOR_YELLOW}[${timestamp}!] ${msg}${RESET_STYLE}" ;;
         error) echo -e "${COLOR_RED}[${timestamp}✗] ${msg}${RESET_STYLE}" >&2 ;;
         question)
+            [[ ${INTERACTIVE_TERMINAL} -eq 0 ]] && return 0
```
BreiteSeite, line 60: This env was not introduced in this PR but do you think the "interactive" check should be updated? Because currently it's misleading/wrong, it checks for being piped instead of testing for interactiveness.

The way bash documents to check for interactiveness is to check for `i` in `$-`. See https://www.gnu.org/software/bash/manual/bash.html#Is-this-Shell-Interactive_003f-1

docs/getting-started/installation.md
```diff
@@ -4,27 +4,72 @@ parent: Getting Started
 nav_order: 2
 ---
 
-# Installation Methods
+# Installing, Updating and Uninstalling
 
 ## Automatic Installation
 
-### One-Line Installation
+### User local Installation (Recommended)
```
BreiteSeite, line 11: Nice!

Not necessary to change but IMHO: the sys-wide install could be a bit reduced in visual importance. Maybe i just write a sentence like "Just run install.sh as root" instead of offering a ready-to-copy-and-paste command. This makes it less likely that people just scan the readme for some command that looks like installation (without reading the surrounding text) and pipe it into a shell. As a bonus: any change on the installation command (url, curl parameter, etc) only need to be updated once and one of the commands has less chance to become "outdated".

src/zwift.sh
```diff
@@ -232,7 +236,12 @@ check_script_up_to_date() {
 }
 
 upgrade_script() {
+    installed_as_root() {
+        [[ ${SCRIPT_DIR} == "/usr/local/bin" ]]
```
BreiteSeite: Would a check if `-O` for the current called script returns false (and `$EUID` is not already `0`) be more robust here instead of assuming the scripts initial installation-location and infer the permissions based on that?

BreiteSeite (changes requested):
Thank you so much for working on my issue! I found 2 things where i think changing makes sense and one opinionated suggestions for the README. :)

docs/getting-started/installation.md
glennvl: Updated in fef88f2ab2323adcd770a0b0d3880329375b553c. I also added a migration guide in 2ae6772dcc2e7b9667a4c0a2cd70be6f2fa80242.

src/zwift.sh
glennvl: We do have full control of the install location, so the original check works. But I agree that your solution is a bit cleaner. Changed in 2ec302c4f8936b84940a99de9dce70ca06091921.

src/zwift.sh
glennvl: The two ways to check described there `[[ ${-} == *i* ]]` and `[[ -n ${PS1:-} ]]` only work when directly executed in bash, they don't work in a script. What does work in a script is `tty -s`, but I'm not sure if that would be better than the current `[[ -t 1 ]]` check.

BreiteSeite: Thanks, sorry i didn't see that this manual page was in a chapter for bash startup files. :( my bad, sorry for noise.

glennvl: We both learned something here, so all good. :+1:

netbrain (approved):

PR #358, opened 2026-05-22
Update documentation

glennvl:
## Summary

Some documentation updates.

- Remove obsolete documentation
- Expand other documentation


docs/advanced/screensaver.md
```diff
@@ -8,3 +8,7 @@ nav_order: 4
 
 If `dbus` is available through a unix socket, the screensaver will be inhibited every 30 seconds to prevent `xscreensaver` or
 other programs listening on the bus from inhibiting the screen.
```
glennvl, line 10: Is this still the case? Or should we remove these lines?

netbrain: I think so, or did we remove dbus check?

netbrain (approved):

docs/advanced/screensaver.md
glennvl: I'm not sure, I've never seen anything related to dbus in any scripts that run inside the container.

PR #368, opened 2026-07-22
Fix star history chart with sealed_token embed

### LINKED ISSUE #367: Star history chart broken
## Describe the bug

Image

## To Reproduce

1. Look at the REAME.md

netbrain:
## Background
GitHub now requires authentication on the stargazers endpoint. Anonymous `GET /repos/netbrain/zwift/stargazers` returns HTTP 401, so the old anonymous star-history embed returns HTTP 500 and no longer renders in the README.

## Change
Switch the star history section to star-history's `sealed_token` embed. The sealed token is an encrypted wrapper around a metadata-read-only token; the raw PAT is not exposed and GitHub secret scanning does not match it, so it is safe to commit in a public README.

## Note
Rendering depends on star-history's service. During testing it intermittently returned `timeout of 10000ms exceeded` (server-side, not a token error). Confirm the chart renders before relying on the merge.

Closes #367

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,15 @@ and tricks :smiley:
 
 ## ⭐ Star History (for fun and giggles)
 
-[![Star History Chart](https://api.star-history.com/svg?repos=netbrain/zwift&type=Date)](https://star-history.com/#netbrain/zwift&Date)
+<!-- markdownlint-disable line-length -->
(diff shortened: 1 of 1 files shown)


glennvl:
It's working again now. Maybe just keep it as is for a while? If it turns out to be down most of the time, we can remove it later. Maybe in the meantime star-history also comes up with a more reliable solution.

PR #369, opened 2026-07-22
Remove broken star history chart

### LINKED ISSUE #367: Star history chart broken

netbrain:
## Background
The star history chart depends on the GitHub stargazers endpoint, which now requires authentication (anonymous returns HTTP 401). PR #368 switched the embed to a star-history `sealed_token`, but it still returns HTTP 500 from star-history's backend, so the chart does not render.

## Change
Remove the Star History section from the README. A reliably rendering chart would need a CI-generated static image, which is not worth it for a for-fun section.

Closes #367

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,12 +114,6 @@ and tricks :smiley:
     cake, and have it too by creating `vgpus` for vms that leverage the host
     gpu, no dedicated gpu required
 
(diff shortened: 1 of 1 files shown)


netbrain:
Looks like it's working now.

PR #372, opened 2026-08-06
Fixes Star History

### LINKED ISSUE #367: Star history chart broken

Mubelotix:
Fixes #367

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ and tricks :smiley:
 ## ⭐ Star History (for fun and giggles)
 
 <!-- markdownlint-disable line-length -->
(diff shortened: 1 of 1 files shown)


glennvl:
image

Looks suspicious to me...

Mubelotix:
I can't force you to trust me but I have a pretty good track record if you dig more

netbrain:
actually looks legit!

netbrain:
Thanks for the effort here, appreciated. After some consideration we've decided not to point the README at a third-party chart service, and instead remove the chart entirely until the official one works again without workarounds. Closing in favor of removal.

Mubelotix:
> the official one

I see what you mean, but I want to remind everyone that there is no officiality about it to be honest. None is endorsed by github, they really do not care about us

PR #389, opened 2026-09-07
Fix container startup on rootless podman 5.x, add ZWIFT_RIDER profile selector

### LINKED ISSUE #285: Extremely low framerate after latest update (WINE_EXPERIMENTAL_WAYLAND=1)

### LINKED ISSUE #388: Rootless podman 5.x: forced --privileged and $USER-based naming break container startup
## Summary

On a host that upgraded from podman 4.x to podman 5.x (rootless), `zwift.sh` fails to start the container. Two independent problems in the script surface, both rooted in podman 5.x behaviour changes. The same script + image worked before the upgrade. Filing with root cause and the workarounds that fixed each.

Environment: NixOS, rootless podman 5.8.4, crun, nvidia legacy 390 via CDI (`--device=nvidia.com/gpu=all`), wine (32-bit Zwift).

## Problem 1: forced `--privileged` breaks rootless podman 5.x

On a non-SELinux host the script always adds `--privileged --security-opt label=disable`:

```
else
    msgbox warning "Not using SELinux, running container in privileged mode to be able to access the GPU"
    container_args+=(--privileged --security-opt label=disable)
fi
```

Under rootless podman 5.x, `--privileged` makes podman build `/dev` itself, which a rootless user namespace cannot do:

- crun: `mount 'devpts' to 'dev/pts': Invalid argument`
- runc: `create device inode /dev/dri/card0: openat card0: permission denied`

The privileged rationale is "to be able to access the GPU", but with CDI (`--device=nvidia.com/gpu=all`) the GPU is available without privileged. Dropping `--privileged` (keeping `--security-opt label=disable`) makes the container start and the game run.

There is no way to disable privileged from the outside: `PRIVILEGED_CONTAINER=1` only forces it on, the SELinux branch is the only non-privileged path, and `CONTAINER_EXTRA_ARGS="--privileged=false"` does not fully undo it (devpts still fails).

Suggested fix: don't force `--privileged` on non-SELinux hosts, or add an env toggle (e.g. `PRIVILEGED_CONTAINER=0`) that actually yields a non-privileged run.

## Problem 2: rider identity keyed off `$USER` breaks single-login multi-rider

The container name, data volume and per-rider config are derived from `$USER`:

```
--name "zwift-${USER}"
-v "zwift-${USER}:${ZWIFT_VOLUME}"
load_config_file "${USER_CONFIG_DIR}/${USER}-config"
```

To run several riders from one shared OS login (a kiosk PC where two people cycle), the natural move is `USER=elin zwift`. But under rootless podman 5.x, crun looks up `$USER` for the container user's group setup. If `elin` has no passwd entry, group setup fails and the container aborts (`setgroups: Invalid argument`, and downstream the same devpts error). Overriding `$USER` to any name without a real account breaks the run.

Suggested fix: introduce a dedicated rider selector, e.g. `ZWIFT_RIDER`, and use `${ZWIFT_RIDER:-$USER}` for the container name, volume and config lookup. That keeps `$USER` as the real login user (which podman/crun need) while allowing multiple rider profiles under one account.

## Also needed for rootless podman 5.x

`--group-add keep-groups` was required so crun could set the container user's supplementary groups at creation; without it, `setgroups: Invalid argument` at container start.

## Verified fix

Local patch: strip `--privileged` from the two `container_args` lines, key the name/volume/config off `${ZWIFT_RIDER:-$USER}`, and pass `CONTAINER_EXTRA_ARGS="--group-add keep-groups"`. Result: container starts, wine runs, ZwiftApp authenticates and renders at 60 FPS on nvidia.

netbrain:
Fixes #388.

## What changed

- **Non-SELinux hosts no longer get `--privileged`.** The else branch now adds only `--security-opt label=disable`. Under rootless podman 5.x, `--privileged` makes podman build `/dev` itself, which a rootless user namespace cannot do (crun: `mount 'devpts' to 'dev/pts': Invalid argument`; runc: `create device inode /dev/dri/card0: permission denied`). GPU access does not need privileged mode: it comes from CDI (`--device=nvidia.com/gpu=all`), `--gpus=all` or `--device=/dev/dri`. `PRIVILEGED_CONTAINER=1` still forces privileged mode as an escape hatch.
- **`--group-add keep-groups` is added automatically on the podman path.** Keeps the host's supplementary groups (video/render) so group-gated `/dev/dri` nodes stay accessible without privileged mode, and avoids the `setgroups: Invalid argument` failure seen on rootless podman 5.x. crun honors the flag, runc silently ignores it, docker never sees it.
- **New `ZWIFT_RIDER` env var selects the rider profile.** Container name, data volume, `<rider>-config` and `<rider>-graphics.txt` are now keyed off `${ZWIFT_RIDER:-$USER}`. Overriding `USER` no longer works on rootless podman 5.x because podman/crun resolve `$USER` to a real account; `ZWIFT_RIDER=alice zwift` keeps `$USER` intact while using the `zwift-alice` volume and `alice-config`. This enables multiple riders on one shared login (kiosk PC).
- Docs updated: `ZWIFT_RIDER` section (replaces the `USER` section), `PRIVILEGED_CONTAINER` description, graphics config path. `crun` added to the cspell word list.

## Why the privileged default is safe to drop

`--privileged` was introduced in #336 to fix #285 (low framerate on Ubuntu). The actual culprit there was `--security-opt label=type:container_runtime_t` on a non-SELinux host, not the lack of privileges. The new default (`label=disable` only) avoids that flag too, so #285 does not come back.

## Backwards compatibility

- `ZWIFT_RIDER` defaults to `$USER`, so existing volumes/config names are unchanged.
- `USER=fred zwift` still selects the `fred` profile where the container tool allows it.
- Docker path is untouched apart from losing `--privileged` (docker GPU access goes through `--gpus=all`/`--device`, which do not need it).
- The nix flake's `privilegedContainer` option keeps its semantics.

## Tested

On NixOS with rootless podman 5.7.0 + crun (this repo's shellcheck/shfmt/markdownlint/cspell all pass):

- `DRYRUN=1` default podman path: no `--privileged`, `--group-add keep-groups` and `--security-opt label=disable` present.
- `DRYRUN=1 ZWIFT_RIDER=alice`: loads `alice-config`, uses `zwift-alice` name/volume, `--userns keep-id` unchanged.
- `DRYRUN=1 CONTAINER_TOOL=docker`: no `keep-groups`, no `--privileged`.
- `DRYRUN=1 PRIVILEGED_CONTAINER=1`: `--privileged` restored.
- Simulated `getenforce`=Enforcing: SELinux branch unchanged (`label=type:container_runtime_t`).
- Live container runs with `--userns keep-id:uid=1000,gid=1000 --group-add keep-groups --security-opt label=disable --device=/dev/dri:/dev/dri`: starts fine, host supplementary groups preserved, `/dev/dri/*` read/write OK.
- Full GPU rendering with this exact set of changes was verified by the reporter of #388 (ZwiftApp authenticates and renders at 60 FPS on nvidia via CDI).

https://claude.ai/code/session_01Ru7So82PdCtPubA12VjuyS


glennvl (approved):
Probably much safer to have a dedicated variable ZWIFT_RIDER instead of overwriting USER indeed. The removal of the privileged flag in the default case is also a good thing in my opinion, it's been a point of discussion a few times already.

PR #396, opened 2026-09-23
Add a custom Wine build for Bluetooth testing

purm:
Wine reaches Bluetooth Low Energy devices through BlueZ, so add a way to compile and test Wine changes without touching the published image:

- build-custom-wine.sh clones the Wine sources, builds them in a Debian builder, and layers the installation over the base image. The sources come from the PR-BT-BLE branch of the wine-with-ble repository (WINE_REPO/WINE_REF), so a branch, tag, or commit can be tested without a local checkout.
- build-ble-wine.sh and launch-ble-wine.sh build and run the Bluetooth test image with focused WINEDEBUG channels and one persistent log per run, named after the Wine commit baked into the image.
- zwift.sh forwards the host D-Bus system bus socket so the winebth driver can reach BlueZ, and warns when the socket is missing.
- The packaged Wine is updated to devel 11.17, which carries the current BlueZ/Bluetooth work.


purm:
PR is not meant to be merged..

